PR #67, opened 2015-11-28
Removing unnecessary JS call.

BeigeBadger:


shakeelmohamed:
We may want to do the same thing for $("#playerTime").show();

BeigeBadger:
It looks like it already works like this?

shakeelmohamed:
Well, almost. I can actually solve this with CSS - I'll make a PR for that soon :)

PR #69, opened 2015-12-02
Adds support for making URLs into hyperlinks in the video description.

### LINKED ISSUE #66: Hyperlink video description contents
If we add hyperlinks to any URLs in the video description the experience will be much better.

The links should follow the form: `<a href="http://example.com" target="_blank">` so the link opens in a new window/tab.

eirism:
Currently the RegEx will only recognize TLD of two or three characters. It covers most cases, but not all.
Fixes #66

### DIFF
--- a/js/everything.js
+++ b/js/everything.js
@@ -272,20 +272,24 @@ function getVideoDescription(videoID) {
             return;
         }
         var description = data.items[0].snippet.description;
-        /* RegEx to match http or https addresses
-        * This will currently only match TLD of two or three letters
-        * Ends capture when:
-        *    (1) it encounters a TLD
-        *    (2) it encounters a period (.) or whitespace, if the TLD was followed by a forwardslash (/) */
-        var re = /((?:http|https)\:\/\/[a-zA-Z0-9\-\.]+.[a-zA-Z]{2,3}(?:\/\S*[^\.\s])?)/g;
-        /* Wraps all found URLs in <a> tags */
-        description = description.replace(re, '<a href="$1" target="_blank">$1</a>');
+        description = highLightURL(description);
         $("#zen-video-description").text(description);
     }).fail(function(jqXHR, textStatus, errorThrown) {
         var responseText = JSON.parse(jqXHR.error().responseText);
         hasError = true;
         showErrorMessage(responseText.error.errors[0].message);
     });
+}
+
+function highLightURL(text) {
+    /* RegEx to match http or https addresses
+    * This will currently only match TLD of two or three letters
+    * Ends capture when:
+    *    (1) it encounters a TLD
+    *    (2) it encounters a period (.) or whitespace, if the TLD was followed by a forwardslash (/) */
+    var re = /((?:http|https)\:\/\/[a-zA-Z0-9\-\.]+.[a-zA-Z]{2,3}(?:\/\S*[^\.\s])?)/g;
+    /* Wraps all found URLs in <a> tags */
+    return text.replace(re, '<a href="$1" target="_blank">$1</a>');
 }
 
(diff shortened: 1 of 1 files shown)


shakeelmohamed:
Can we pull this logic out into a function? It will make future refactor, if any, easier

eirism:
Is this better? I know that there are URLs that won't be recognized by the pattern, but it will at least hyperlink the most common forms.

js/everything.js
```diff
@@ -271,7 +271,9 @@ function getVideoDescription(videoID) {
             showErrorMessage("Video description not found");
             return;
         }
-        $("#zen-video-description").text(data.items[0].snippet.description);
+        var description = data.items[0].snippet.description;
+        description = highLightURL(description);
+        $("#zen-video-description").text(description);
```
shakeelmohamed: We'll need to change `.text()` to `.html()` to to properly render the HTML

js/everything.js
```diff
@@ -279,6 +281,17 @@ function getVideoDescription(videoID) {
     });
 }
 
+function highLightURL(text) {
+    /* RegEx to match http or https addresses
+    * This will currently only match TLD of two or three letters
+    * Ends capture when:
+    *    (1) it encounters a TLD
+    *    (2) it encounters a period (.) or whitespace, if the TLD was followed by a forwardslash (/) */
+    var re = /((?:http|https)\:\/\/[a-zA-Z0-9\-\.]+.[a-zA-Z]{2,3}(?:\/\S*[^\.\s])?)/g;
```
shakeelmohamed: This doesn't appear to work as is for our "demo" video description. If I change the last `?` to `+` it works fine - thoughts?

You can try it on the following text:
```
OUT NOW! Download the new EP "Piece Of The Indestructible" here: http://radi.al/Indestructible

TOUR DATES: http://theglitchmob.com/tour
9/18-20 - Sacramento, CA - TBD Festival
12/31 - Birmingham, Alabama - Bassnectar NYE 360

Vinyl. Merchandise and more here: http://smarturl.it/TGMstore

The Glitch Mob
http://fb.me/theglitchmobmusic
http://twitter.com/theglitchmob
http://instagram.com/theglitchmob
http://soundcloud.com/theglitchmob
http://youtube.com/glitchmob
SNAPCHAT: theglitchmob
```

Before:

capturfiles_5

After:

capturfiles_6

shakeelmohamed:
Thanks @eirism, just a couple small things then I'm happy to merge this in!

If you need to test out the video description functionality, you can copy/symlink this project's files onto a web server (ie: MAMP, XAMPP, WAMP) then access them via `localhost`. Our tests are not yet able to verify this kind of thing :smile:

js/everything.js
```diff
@@ -279,6 +281,17 @@ function getVideoDescription(videoID) {
     });
 }
 
+function highLightURL(text) {
```
shakeelmohamed: nit: I'd prefer a better function name, maybe something like `anchorURLs` or `stringToAnchoredHTML`

eirism:
I believe the mismatching is because of an missing escape character, it did not escape the period before the TLD. Did you test it on a web server? When I tested it locally with the fix, the description is not parsed as HTML, and therefore the <a> tags are outputted as plain text.
To fix this we might have to write to the DOM instead of document.write

eirism:
This question on stackowerflow, http://stackoverflow.com/questions/1960461/convert-plain-text-urls-into-html-hyperlinks-in-php, concerns the recognition of URLs. It seems like their solutions covers more or less the same as mine.

shakeelmohamed:
This is looking great, thanks so much for hopping on this @eirism!

PR #92, opened 2015-12-29
Reduce/remove timeout on play/pause click

### LINKED ISSUE #74: Reduce/remove timeout on play/pause click
The current timeout of 1 second makes the button seem unresponsive. And if you click the button during the timeout, the click won't be registered.

We can either experiment with setting the timeout lower or implement some logic to switch between the play and pause button some other way (using setInterval and PlayerState?) .

monicacheung:
Closes #74

### DIFF
--- a/js/everything.js
+++ b/js/everything.js
@@ -22,9 +22,15 @@ function onYouTubeIframeAPIReady() {/* jshint ignore:line */
                 //console.log("State changed to " + event.data);
                 var playerState = event.data;
 
-                 if (playerState === YT.PlayerState.ENDED) {
+                if (playerState === YT.PlayerState.ENDED) {
                     showPlayButton();
-                 }
+                }
+                else if (playerState === YT.PlayerState.PAUSED) {
+                    showPlayButton();
+                }
+                else if (playerState === YT.PlayerState.PLAYING) {
+                    showPauseButton();
+                }
             },
             "onError": function(event) {
                 var message = "Got an unknown error, check the JS console.";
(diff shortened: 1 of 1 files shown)


js/everything.js
```diff
@@ -22,9 +22,15 @@ function onYouTubeIframeAPIReady() {/* jshint ignore:line */
                 //console.log("State changed to " + event.data);
                 var playerState = event.data;
 
-                 if (playerState === YT.PlayerState.ENDED) {
+                if (playerState === YT.PlayerState.ENDED) {
```
shakeelmohamed: Can we replace this with a switch/case?

monicacheung: Okay, how can I update my diff in the pull request?

shakeelmohamed: You can make another commit, hop in the gitter chat if you need anything :smile:

js/everything.js
```diff
@@ -22,9 +22,17 @@ function onYouTubeIframeAPIReady() {/* jshint ignore:line */
                 //console.log("State changed to " + event.data);
                 var playerState = event.data;
 
-                 if (playerState === YT.PlayerState.ENDED) {
-                    showPlayButton();
-                 }
+                switch (playerState) {
+                    case YT.PlayerState.ENDED:
```
shakeelmohamed: We can do slightly better here using fall-through so we don't have to duplicate code for `ENDED` and `PAUSED`. See this [SO question](http://stackoverflow.com/questions/13207927/switch-statement-multiple-cases-in-javascript#13207939) for an example.

Also, we should probably have a default condition while we're at at it.
We're currently not addressing these cases.
- `YT.PlayerState.BUFFERING`
- `YT.PlayerState.CUED`

Both of which I believe should have `showPlayButton();` as their default behavior. So, maybe we can simplify this whole thing quite a bit - just have 1 case for `YT.PlayerState.PLAYING`, and a default to `showPlayButton();`

Does that make sense?

monicacheung: Yup. Good catch :)

shakeelmohamed:
Thank you @monicacheung! Welcome to the org :rocket:

PR #94, opened 2015-12-29
Remove leading music notes from title. Closes #87

### LINKED ISSUE #87: Don't prefix the title with the music note if it already begins with one
It would be nice to somehow avoid this from happening, note the 2 musical notes at the beginning of the title. The first one is ours, the second is actually in the title!

capturfiles_2

Here is an example video: https://www.youtube.com/watch?v=iXAbte4QXKs

monicacheung:


shakeelmohamed:
This is certainly an option. What if instead we only prepend `"<i class=\"fa fa-music\"></i>"` if the first character in the title is not a music note? My personal preference is to not mess with the video title - who knows what kind of issues could come from that.

There's a few [other music note characters](https://en.wikipedia.org/wiki/Miscellaneous_Symbols#Compact_table) (see row `U+266x`) we may want to consider checking for as well

shakeelmohamed:
Looks great! You'll need to rebase your local branch since another PR has come in since you forked.

On the command line (`origin` is the default remote repository, so yours may be different, run `git remote` to find out):

```shell
git remote add org git@github.com:zen-audio-player/zen-audio-player.github.io.git
git pull --rebase org/master
git push origin master
```

With any luck, that last command should fail, so you can run `git push -f origin master` to do a force push. I have a local copy of your changes in case anything goes wrong :smile:

monicacheung:
Sorry for the mess. :( This is kind of confusing. Does it look okay now?

shakeelmohamed:
@monicacheung No problem at all, this is a very common problem actually. Thanks for doing this!

PR #114, opened 2016-01-05
Migrate from html5lint to HTMLHint - fixes #100

### LINKED ISSUE #100: Replace html5-lint dependency
See #98. The server html5-lint was using has been shut down, so we need to find another way to do HTML validation for the tests.
I am very open to suggestions here!

agauniyal:
Tests pass, updated package.json and included .htmlhintrc


package.json
```diff
@@ -24,7 +26,7 @@
   },
   "homepage": "https://zen-audio-player.github.io",
   "devDependencies": {
-    "html5-lint": "^0.2.5",
+    "htmlhint": "latest",
```
shakeelmohamed: Can we at least stick to the `0.9.x` line? I don't want something crazy to happen if they go to `1.0.0`

.htmlhintrc
```diff
@@ -0,0 +1,23 @@
...
+    "tag-pair": true,
+    "spec-char-escape": true,
+    "id-unique": true,
+    "src-not-empty": true,
+    "attr-no-duplication": true,
+    "title-require": true,
+    "head-script-disabled": false,
+    "alt-require": true,
+    "doctype-html5": true,
+    "id-class-value": 'dash',
```
shakeelmohamed: nit: single quotes

.htmlhintrc
```diff
@@ -0,0 +1,23 @@
...
+    "style-disabled": true,
+    "inline-style-disabled": true,
+    "inline-script-disabled": true,
+    "space-tab-mixed-disabled": true,
+    "id-class-ad-disabled": true,
+    "href-abs-or-rel": 'abs',
```
shakeelmohamed: nit: single quotes

package.json
```diff
@@ -4,7 +4,9 @@
   "description": "Listen to YouTube videos, without the distracting visuals.",
   "main": "tests.js",
   "scripts": {
-    "pretest": "jshint js test",
+    "lint:js": "jshint js test",
+    "lint:html": "htmlhint index.html",
+    "pretest": "npm run lint:js && npm run lint:html",
```
shakeelmohamed: I like the way you broke this up into multiple scripts. For `pretest`, can we do something like `npm run test -- --grep="lint"`. It seems possible as of npm v2.0.0

test/html-check.js
```diff
@@ -1,36 +0,0 @@
...
-                    var lineNumber = msg.lastLine;
-                    console.log("HTML5 Lint [%s]: %s %s (line %s)", type, message, lineNumber);
-                    assert.notStrictEqual(type, "error", message + ", see line " + lineNumber + " in index.html");
-                }
-            }
-            done();
-        });
-    });
-    it("should not contain tab characters", function() {
-        assert.strictEqual(-1, _html.indexOf("\t"), "Please remove any tab indents from index.html");
```
shakeelmohamed: I'm assuming `space-tab-mixed-disabled` will take care of this case

package.json
agauniyal: Yup works that way although I wasn't aware of it :smile:

agauniyal:
I've updated package.json file and had to turn off some checks which were interferring with jquery and absolute/relative url consistencies in this project.

shakeelmohamed:
> I've updated package.json file and had to turn off some checks which were interferring with jquery and absolute/relative url consistencies in this project.
@agauniyal can you make a note of those options and create a new issue for fixing them? Or, are they unfixable?

shakeelmohamed:
Looks good. Thanks for doing this work @agauniyal, welcome to the org!

agauniyal:
`absolute url` rule cannot be fixed and I don't believe is an absolute rule to enforce. There's another variable-naming issue rule with jquery which I'll try after a few days (possibly after adding csslint , eslint PR).

shakeelmohamed:
@agauniyal Great, thanks for the effort!

PR #115, opened 2016-01-05
Fix race condition loading youtube api - Fixes #42

### LINKED ISSUE #42: Demo link does nothing in Microsoft Edge or Internet Explorer
Using Microsoft Edge, the demo link doesn't do anything. The URL clearly changes, but it just dies (could be a race condition with jQuery loading before the YouTube API). There are no JavaScript console errors.

edit: Non-Windows users can get free VMs from Microsoft's website! https://dev.windows.com/en-us/microsoft-edge/tools/vms/

image

moen-io:
This resolves #42 for me, running Microsoft Edge on Windows 10. Appears to be a race condition where the youtube api attempts to call onYouTubeIframeAPIReady before it has been defined. Closes #42.


monicacheung:
Can you give more info about how this is race condition? onYouTubeIframeAPIReady() should only get called after the youtube javascript loads. I think moving the youtube script under everything.js will make the some of the player-related things load before the youtubejs loads. Can you give printout of the player state in onStateChange(), onYouTubeIframeAPIReady(), and print out the errors from onError()?

monicacheung:
Looking into how scripts executed and seems like IE doesn't always execute them in order and might even execute one partially and stop to execute another. So this fix might sometimes give desired result... But in browsers that execute them one after another, it might cause some things we have in everything.js (volume slider, etc) to load before the youtubejs finish.

shakeelmohamed:
@moen-io This seems like a strange solution. I think the correct fix this #42 might involve using [requireJS](http://requirejs.org/)

moen-io:
You are correct that onYouTubeIframeAPIReady is only called after the youtube scripts have finished loading. The issue is that www-widgetapi.js loads before everything.js in Edge. Observe:

Edge
edge

Chrome
chrome

Thus onYouTubeIframeAPIReady is not yet defined when youtube attempts to hook it up. Changing the load order should ensure that everything.js is always fully loaded before the youtube api is fetched.

Everything.js also only sets up the player on document.ready, so changing the load order should not make a difference there.

shakeelmohamed:
@moen-io If you look at the source of `https://www.youtube.com/iframe_api`, it's basically including `www-widgetapi.js`

moen-io:
@shakeelmohamed Yeah that appears to be the source of the problem. It could probably be solved in a number of ways, but at the end of the day it's just a matter of ensuring that the scripts are loaded in the correct order.

shakeelmohamed:
I will leave this open until we try using requireJS

monicacheung:
http://api.jquery.com/jQuery.getScript/

Can you look at this and see if you can use it to load the youtube script and everything.js in order?

shakeelmohamed:
@monicacheung I just tried that with your fork and it works enough to hear the video but it's horribly unusable (unrelated to your changes, I'm guessing). The file took about 10s to load and `$("#audioplayer").show();` was never triggered so there were no controls! There were also constant parameter relater errors in `www-widgetapi.js`.

I did look into requireJS, and it may be overkill for our problem - especially since we're dependent on YouTube's remote JS library.

@moen-io I am closing this for now, as I believe we'll have something better to go off of when #9 is merged in.

Thanks for the effort, and let us know if you think of another solution!

shakeelmohamed:
After some re-consideration and manual testing, this change works great in IE 12, after disabling compatibility settings for local development (see screenshot below). However, in Edge I have yet to see autoplay work correctly. When clicking the demo link, the player gets setup correctly, but it never starts playing.

So, I propose we merge this in immediately after #9 as the changes there should resolve the autoplay issues.

@moen-io If you want/need this done before #9, you can grab some of the changes from @monicacheung's [fork](https://github.com/monicacheung/zen-audio-player.github.io/commit/0dfe6532cf08629ffd64e10b13e71dd031b7e40e) - we resolve the merge conflicts later. Ping me on gitter if you need guidance on that. Additionally, I think we may be able to speed up loading time by moving all `<script>` tags to the bottom of the body - what do you think?

capturfiles_11

moen-io:
@shakeelmohamed Autoplay seemed to work fine for me Edge using Mongoose as a local server. Some problem in `onPlayerReady` perhaps? I'll see if I can replicate your issue this evening.

And yeah merging #9 first seems like a good idea, so feel free to hold off on this until that's sorted out.

shakeelmohamed:
We just merged #120 for #9. Rebase your change and I'm happy to bring it in. I don't think we're quite out of the woods though. Many things appear to be broken visually (with your changes, run over `file://` - I'm not able to get apache running correctly on my Windows machine at the moment so I can't test)

capture1

moen-io:
Yeah it doesn't look like things work properly when run through `file://`. You can use [mongoose](https://www.cesanta.com/products) as a simple web server. Just run the .exe in the project folder and it will be served on localhost:8080. Looks good to me then!

shakeelmohamed:
@moen-io Great, I tried mongoose and it does half the job - I still saw CORS issues for search & getting the description but that's good enough for now. After the merge it's working great on Edge, thank you and welcome to the org!

I've added mongoose to a wiki page for reference https://github.com/zen-audio-player/zen-audio-player.github.io/wiki/How-to-test-locally

PR #119, opened 2016-01-07
Migrate to eslint, fixes #95

### LINKED ISSUE #95: Migrate from jshint -> eslint
We should be able to match our existing jshint configuration with [eslint](https://www.npmjs.com/package/eslint), which appears to be the better solution.

- [ ] update dependencies in `package.json`
- [ ] update the scripts in `package.json`
- [ ] update configuration wherever it lives
- [ ] remove the `.jshintrc` file, why did I ever check that in? :disappointed:

agauniyal:
Updated `package.json` with eslint dependency as well as eslint conf. Removed jshint conf from `package.json` and `.jshintrc`. 
I had to edit `everything.js` to follow correct switch case indentation which jshint was ignoring.


package.json
```diff
@@ -26,25 +28,41 @@
...
+        "jquery": true,
+        "es6": true
+    },
+    "globals": {
+        "YT": true,
+        "ga": true,
+        "twttr": true,
+        "time": true,
+        "require": true,
+        "__dirname": true,
```
shakeelmohamed: Are `__dirname` and `__filename` required in this list? Can you add `"node": true` to the `env` object?

js/everything.js
```diff
@@ -448,12 +448,19 @@ $(function() {
...
 });
 
-/* jshint ignore:start */
+/*eslint-disable */
 // Google Analytics goodness
-(function(i,s,o,g,r,a,m){i["GoogleAnalyticsObject"]=r;i[r]=i[r]||function(){
-(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
-m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
-})(window,document,"script","//www.google-analytics.com/analytics.js","ga");
+(function(i, s, o, g, r, a, m) {
```
shakeelmohamed: Not a fan of having this expanded, but I can live with it.

js/everything.js
```diff
@@ -448,12 +448,19 @@ $(function() {
     });
 });
 
-/* jshint ignore:start */
+/*eslint-disable */
```
shakeelmohamed: Does this line do anything?

js/everything.js
```diff
@@ -448,12 +448,19 @@ $(function() {
...
+    a = s.createElement(o),
+        m = s.getElementsByTagName(o)[0];
+    a.async = 1;
+    a.src = g;
+    m.parentNode.insertBefore(a, m)
+})(window, document, "script", "//www.google-analytics.com/analytics.js", "ga");
 ga("create", "UA-62983413-1", "auto");
 ga("send", "pageview");
-/* jshint ignore:end */
+/*eslint-enable */
```
shakeelmohamed: Does this line do anything?

package.json
```diff
@@ -4,9 +4,11 @@
   "description": "Listen to YouTube videos, without the distracting visuals.",
   "main": "tests.js",
   "scripts": {
-    "lint:js": "jshint js test",
+    "lint:js": "eslint js/**",
+    "lint:tests":"eslint test/**/*.js",
```
shakeelmohamed: This path doesn't look right

package.json
```diff
@@ -4,9 +4,11 @@
   "description": "Listen to YouTube videos, without the distracting visuals.",
   "main": "tests.js",
   "scripts": {
-    "lint:js": "jshint js test",
+    "lint:js": "eslint js/**",
```
shakeelmohamed: Why `**`?

package.json
```diff
@@ -26,25 +28,41 @@
...
-    "indent": 4,
-    "loopfunc": true,
-    "newcap": true,
-    "node": true,
-    "quotmark": "double",
-    "trailing": true,
-    "unused": true,
-    "white": true,
-    "esversion": 6
+  "eslintConfig": {
```
shakeelmohamed: @agauniyal do you think we should move these to their own file?

agauniyal: @shakeelmohamed your call, I saw previous jshint conf was done in package.json so replicated that. We could also move them into seperate eslintrc.json or .yaml file.

package.json
agauniyal: Hmm never occurred to me :+1:

js/everything.js
agauniyal: Disables eslint just for `\\Google Analytics goodness` part. Otherwise numerous rules would be broken and we'll have to expand that block much more.

package.json
shakeelmohamed: I think `eslintrc.json` makes sense since we're doing that for the other linters

package.json
agauniyal: To be honest I don't know why but documentation mentions this :/. I'll correct it to `*.js`.

js/everything.js
shakeelmohamed: In that case, can we revert the expansion somehow?

agauniyal:
Why aren't travis test running? :open_mouth:

agauniyal:
Rebased the PR :smile:

shakeelmohamed:
Ship it!

Issue #26, opened 2015-06-12
Migrate Google Analytics to Keen


shakeelmohamed:
## Priority level 1
- [x] Create keen.io account
- [x] Have keen.io log as much data as GA has currently
- [x] Follow these guides: [media analytics](https://keen.io/guides/media-analytics/), [data enrichment](https://keen.io/docs/api/?javascript#data-enrichment), [data collection](https://keen.io/docs/data-collection/), [data analysis](https://keen.io/docs/data-analysis/)
- [x] Decide between [keen-js](https://github.com/keen/keen-js) & [keen-tracking](https://github.com/keen/keen-tracking.js) - looks like tracking is not ready for GA, keen-js it is!
- [x] Get Keen to log GA-like demographics data
- [ ] Add an event for when the tweet button is clicked to log share stats
- [ ] Remove GA
- [ ] Find a way to export GA to keen.io (maybe follow [Moz's guide](https://moz.com/blog/how-to-export-google-analytics-data-to-excel-via-the-api))
## Priority level 2
- [X] Create a new repo for Keen dashboards with a `gh-pages` branch: https://github.com/zen-audio-player/stats
- [x] [Example dashboard](http://keen.github.io/dashboards/examples/connected-devices/) more [here](http://keen.github.io/dashboards/examples/)
- [x] [data viz](https://keen.io/docs/data-visualization/) tutorial
- [ ] Start building keen.io dashboards - see current progress [here](http://zen-audio-player.github.io/dashboards/examples/connected-devices/)
- [x] Add table for showing popular songs/artists by parsing video titles
- [ ] Add table for showing common errors
- [ ] chart for search vs. video ID vs. URL
- [ ] Video duration chart

I've built a really nice GA dashboard, it would be really cool to show these stats publicly via [keen.io](https://keen.io/).

![capture](https://cloud.githubusercontent.com/assets/1480063/8124646/b341f068-1092-11e5-99ca-70fc08f6e543.PNG)

shakeelmohamed:
This could be helpful: https://github.com/onefrankguy/frankmitchell.org/blob/master/drafts/replacing-google-analytics-with-keen-io.md

shakeelmohamed:
Made the repo!

shakeelmohamed:
Made tons of progress on the feature/keen.io branch, sent their team a follow-up email for next steps :clock10:

shakeelmohamed:
[...]

PR #124, opened 2016-01-09
Add Keen.io event tracking

shakeelmohamed:
Takes a big step towards #26!
I need to work on the dashboards a bit to see if we're collecting enough data, this is a work in progress


js/everything.js
```diff
@@ -81,10 +132,20 @@ function onPlayerReady(event) {
         return;
     }
 
+    // Google Analytics
```
shakeelmohamed: This change probably needs to be reverted

shakeelmohamed:
The dashboards are not ready yet, but I'm merging this in so we can start capturing real data

PR #133, opened 2016-01-15
Makes request for search auto completion go over https

eirism:
Browsers will block scripts from making http request when a site is accessed via https. Therefore the auto completion did not work on https. This fixes it.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -3,6 +3,8 @@ var Browser = require("zombie");
 var path = require("path");
 var fs = require("fs");
 
+var _js = "";
+
 const browser = new Browser();
 
 var indexHTMLURL = "file://" + path.join(__dirname, "..", "index.html");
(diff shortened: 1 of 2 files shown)


shakeelmohamed:
@monicacheung  - any objections?

test/index.js
```diff
@@ -117,6 +123,9 @@ describe("JavaScript components", function() {
     it("should load ZenPlayer from everything.js", function() {
         assert.ok(browser.evaluate("ZenPlayer"));
     });
+    it("should make all requests over https, not http", function() {
+        assert.strictEqual(-1, _js.indexOf("http://", "Please use HTTPS for all scripts"));
```
shakeelmohamed: There is a `)` out of place here. There 3 arguments for `assert.strictEqual()` should be:

1. `-1`
2. `_js.indexOf("http://")`
3. `"Please use HTTPS for all scripts"`

shakeelmohamed:
Thank you!

PR #135, opened 2016-01-15
Fix google analytics spam

### LINKED ISSUE #134: Google analytics sends events constantly
GA events are sent several times per second as seen here:

image

The root cause is that analytics are sent every time `parseYoutubeVideoID` is called. As it happens, this occurs every 50ms via `updatePlayerTime` and `storeTime`.

I will be submitting a PR shortly with a proposed solution.

moen-io:
This closes #134. GA events are sent constantly every time we access the video id. I have added the current video id as a global variable to ensure this lookup only happens once.


shakeelmohamed:
Ship it!

PR #140, opened 2016-01-19
Added CNAME file

ecnepsnai:


shakeelmohamed:
Great, we should get the [DNS configured](https://help.github.com/articles/setting-up-a-custom-domain-with-github-pages/) first right?

PR #141, opened 2016-01-20
Refactor to common (work in progress)

shakeelmohamed:
**This is very much a work in progress, but I wanted to get the bulk of it started.**

When this is actually done, `js/zap-common.js` will be moved into a bower module and end up moving to something like: `bower_components/zap-common/index.js`

Looking for some feedback here @zen-audio-player/contributors :smiley:

### DIFF
--- a/js/zap-common.js
+++ b/js/zap-common.js
@@ -1,0 +1,64 @@
+/*eslint-disable no-unused-vars*/
+function getParameterByName(url, name) {
+    name = name.replace(/[\[]/, "\\[").replace(/[\]]/, "\\]");
+    var regex = new RegExp("[\\?&]" + name + "=([^&#]*)"),
+        results = regex.exec(url);
+    return results === null ? "" : decodeURIComponent(results[1].replace(/\+/g, " "));
+}
+
+function getSearchResults(query, youTubeDataApiKey, onData, onFail) {
+    $.getJSON("https://www.googleapis.com/youtube/v3/search", {
+        key: youTubeDataApiKey,
+        part: "snippet",
+        q: query,
+        type: "video"
+    }, onData).fail(onFail);
+}
+
+function getAutocompleteSuggestions(query, callback) {
+    $.getJSON("https://suggestqueries.google.com/complete/search?callback=?", {
+        q: query,
+        client: "youtube",
+        ds: "yt"
+    }, callback);
+}
+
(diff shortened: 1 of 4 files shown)


js/zap-common.js
```diff
@@ -0,0 +1,64 @@
...
+    $.getJSON("https://www.googleapis.com/youtube/v3/search", {
+        key: youTubeDataApiKey,
+        part: "snippet",
+        q: query,
+        type: "video"
+    }, onData).fail(onFail);
+}
+
+function getAutocompleteSuggestions(query, callback) {
+    $.getJSON("https://suggestqueries.google.com/complete/search?callback=?", {
```
shakeelmohamed: @monicacheung this was from bb3b87c87b5b751ff2738a0a87394d126128db48, why is `callback=?` in the URL? Does it have any effect?

monicacheung:
@shakeelmohamed To use jsonp for cross-domain request. Without it, we'll see origin not allowed err.

monicacheung:
On a side note, we can get rid of the file:// running site locally err if in the google dev console, you remove whatever's in "Accept requests from these HTTP referrers (web sites) (Optional)".

js/zap-common.js
monicacheung: @shakeelmohamed To use jsonp for cross-domain request. Without it, we'll see origin not allowed err.

js/zap-common.js
```diff
@@ -0,0 +1,64 @@
...
+        // We found a slash in the video ID (ex: real id is ABC123, but saw ABC123/zen)
+        // So, only keep what's before the slash
+        if (slashPos !== -1) {
+            videoInfo.id = videoInfo.id.substring(0, slashPos);
+        }
+
+        return videoInfo;
+    }
+}
+/*eslint-enable */
```
monicacheung: getVideoDescription() should go in this file too.

shakeelmohamed: done, although I'm not sure it will be used by another project

shakeelmohamed:
>On a side note, we can get rid of the file:// running site locally err if in the google dev console, you remove whatever's in "Accept requests from these HTTP referrers (web sites) (Optional)".

@monicacheung Yeah, but the restrictions are there to prevent abuse. I can't justify removing them at this time, since it's only going to affect contributors.

PR #147, opened 2016-02-07
Implements the fix from #146. Closes #146

### LINKED ISSUE #146: If searching with a string containing &, the right side will be ignored
This is likely happening somewhere in `$("#form").submit()` in `js/everything.js`, possibly more than just 1 place.

A trivial solution could be replace " & " with "and"

eirism:


shakeelmohamed:
Ship it! Thanks for digging into this issue

PR #150, opened 2016-02-14
Possible fix for using app on Android - Issue #5

BlinkBP:
I checked in Chrome on my smartphone. It says "ERROR: Invalid YouTube videoID or URL".
The same thing happens in Windows version of Chrome with Device mode enabled. From a little bit of debugging I found out that the video gets cued (playerState = 5) but the app is only checking whether video is buffering (playerState = 3). If I order the player to play a video it plays fine.

### DIFF
--- a/js/everything.js
+++ b/js/everything.js
@@ -75,10 +75,8 @@ function onPlayerReady(event) {
     updateTweetMessage();
 
     // If the video isn't going to play, then return.
-    if (event.target.getPlayerState() == YT.PlayerState.UNSTARTED) {
-        if (currentVideoID.length > 0) {
-            errorMessage.show("Invalid YouTube videoID or URL.");
-        }
+    if (event.target.getPlayerState() === YT.PlayerState.UNSTARTED) {
+        errorMessage.show("Invalid YouTube videoID or URL.");
         return;
     }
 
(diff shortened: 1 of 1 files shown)


shakeelmohamed:
Use `===` or the tests will not pass. Can we check for either condition? I don't remember the reason checking for the `BUFFERING` state, but I'd rather not introduce another bug :smile:

BlinkBP:
Yeah. I don't really know either. Also, this "if" statement checking if a video has length > 0. Why would one check if a video has length if the URL or video id is bad?

shakeelmohamed:
If the state is `PLAYING`, we'd be showing an error when there isn't one. What if we just change the condition to be `event.target.getPlayerState() < 0`? This seems like the right fix according to [the docs](https://developers.google.com/youtube/iframe_api_reference#Playback_status)

BlinkBP:
I don't fully understand what do you mean. `YT.PlayerState.UNSTARTED` always equals -1 and that's the only state we don't want. `event.target.getPlayerState() < 0` is basically the same solution in this situation.

shakeelmohamed:
@BlinkBP Got it, my mistake I was misreading :+1:

Issue #136, opened 2016-01-15
Add support for plyr


shakeelmohamed:
There's a really well designed HTML5 audio player called [audio.js](https://github.com/kolber/audiojs). Are we able to override the default functionality of all the controls to behave like our current controls? I'm also open to alternative libraries - what do we think? @zen-audio-player/contributors 

<img width="481" alt="capturfiles_3" src="https://cloud.githubusercontent.com/assets/1480063/12368125/78b3a7fa-bb9d-11e5-8687-61f0def9cea2.png">

One reason to use this is when #1 (playlists) is resolved, we can have an [awesome UI](http://kolber.github.io/audiojs/demos/test6.html).

<img width="571" alt="capturfiles_2" src="https://cloud.githubusercontent.com/assets/1480063/12368127/7d6fe1fa-bb9d-11e5-953b-88518f57662e.png">

agauniyal:
Latest commit 7af47a1 on Dec 11, 2013
100+ Issues
15 PR's not merged - back from 2012

:open_mouth:

shakeelmohamed:
@agauniyal Yeah, but it works really well! If it works, I'm not concerned about it being maintained

shakeelmohamed:
Here's an alternative:

https://plyr.io/ - https://github.com/selz/plyr
![capturfiles_1](https://cloud.githubusercontent.com/assets/1480063/12374649/eb855f62-bc56-11e5-8bfa-6f8de7ec6753.png)

They actually have a YouTube player that we could use with just 1 line of css (to hide the video).

### After

![capturfiles_2](https://cloud.githubusercontent.com/assets/1480063/12374692/6a1536a8-bc58-11e5-9e79-a877fba6adf7.png)


### Before
![image](https://cloud.githubusercontent.com/assets/1480063/12374690/5952d7da-bc58-11e5-8d2c-6e3cefb0e575.png)

monicacheung:
Plyr looks nice. It says it's customizable too... Who wants to play around with this? :laughing:

joshghent:
I'm looking to contribute to this project so I'll take a look.

shakeelmohamed:
Thanks @joshghent

PR #153, opened 2016-02-29
#136 - Plyr player as a base player

BlinkBP:
Added Plyr scripts to `index.html` (js script + svg for icons) (soundcloud is there by mistake; i will delete it in the next commit).  
Added code to `everything.js` to use Plyr with current video id.  
**Plyr works independently to current player and only sits next to it at the moment.  **
I think we should use Plyr instead of current player. It will be easy to add Soundcloud support (already tried and works good) and also YT playlists (but I'm not sure of it as of now).


index.html
```diff
@@ -25,6 +26,9 @@
     <script src="js/zap-common.js"></script>
     <script src="js/everything.js"></script>
     <script src="https://www.youtube.com/iframe_api"></script>
+	<script src="https://connect.soundcloud.com/sdk/sdk-3.0.0.js"></script>
```
shakeelmohamed: Let's remove soundcloud here, that should be added in it's own PR

index.html
```diff
@@ -25,6 +26,9 @@
...
+	<script src="https://cdn.plyr.io/1.5.16/plyr.js"></script>
```
shakeelmohamed: We should be using bower. `bower install --save plyr`, then the `src` should look more like the other bower components above. While we're at it, this line should be moved up a bit so it's grouped with the others

js/everything.js
```diff
@@ -84,6 +85,24 @@ function onPlayerReady(event) {
...
+			plyrPlayer.source({
+				type: 'video',
+				title: player.getVideoData().title,
+				sources: [{
+					src: currentVideoID,
+					type: 'youtube'
+				}]
+			});
+			//Hide video; leave only controls
+			$(document.querySelector('.plyr__video-wrapper')).hide();
```
shakeelmohamed: This should just be `$(".plyr__video-wrapper").hide();`

index.html
```diff
@@ -117,5 +123,44 @@
             "logo": "https://zenplayer.audio/img/zen-audio-player-905.png"
         }
     </script>
+	<!-- Load SVG defs for Plyr-->
```
shakeelmohamed: Why do we need to do all of this?

index.html
```diff
@@ -16,15 +16,18 @@
     <link rel="stylesheet" href="bower_components/font-awesome/css/font-awesome.min.css">
     <link rel="stylesheet" href="bower_components/seiyria-bootstrap-slider/dist/css/bootstrap-slider.min.css">
     <link rel="stylesheet" href="css/styles.css">
-    
+    <link rel="stylesheet" href="https://cdn.plyr.io/1.5.16/plyr.css">
```
shakeelmohamed: Don't forget to update this when switching to bower

BlinkBP:
Will do all tomorrow. I should already be asleep for tomorrow classes. ;)

index.html
BlinkBP: This loads SVG's for control's icons. They (Plyr github) suggest to use something like gulp-svg-sprite instead of this.

BlinkBP:
Travis throws `89:22  error    "plyr" is not defined `. Do we really need something like requireJS to shut Travis up?

shakeelmohamed:
@BlinkBP you can fix that by setting plyr as a global for eslint, the following patch will do that (tested locally)

```diff
diff --git a/.eslintrc.json b/.eslintrc.json
index 9315d7f..f554dec 100644
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -11,7 +11,8 @@
         "YT": true,
         "ga": true,
         "twttr": true,
-        "time": true
+        "time": true,
+        "plyr": true
     },
     "rules": {
         "indent": [

```

.eslintrc.json
```diff
@@ -11,7 +11,8 @@
         "YT": true,
         "ga": true,
         "twttr": true,
-        "time": true
+        "time": true,
+		"plyr": true
```
shakeelmohamed: nit: tab indent

BlinkBP: Won't happen again. Already changed my editor's prefeneces. :smiley:

shakeelmohamed: Great :smile:

index.html
```diff
@@ -8,25 +8,29 @@
...
     <script type="text/javascript">window._trackJs = { token: "f58d0b7c4b4248939fddc582f9e55c9d" };</script>
     <script src="bower_components/trackjs/tracker.js"></script>
     <script src="bower_components/jquery/dist/jquery.min.js"></script>
     <script src="bower_components/seiyria-bootstrap-slider/dist/bootstrap-slider.min.js"></script>
     <script src="bower_components/typeahead.js/dist/typeahead.jquery.min.js"></script>
+    <script src="bower_components/plyr/dist/plyr.js"></script>
+    <script src="https://www.youtube.com/iframe_api"></script>
     <script src="js/zap-common.js"></script>
     <script src="js/everything.js"></script>
-    <script src="https://www.youtube.com/iframe_api"></script>
```
shakeelmohamed: Can we revert this? We're going to run into #115 again

BlinkBP: Good catch.

js/everything.js
```diff
@@ -84,6 +85,33 @@ function onPlayerReady(event) {
...
+            type: "video",
+            title: player.getVideoData().title,
+            sources: [{
+                src: currentVideoID,
+                type: "youtube"
+            }]
+        });
+        //Hide video; leave only controls
+        $(".plyr__video-wrapper").hide();
+        //Inject svg with controls' icons
```
shakeelmohamed: jQuery actually have a `load()` function, it's deprecated but I think it looks better than their hacky method

```javascript
$("#plyr-svg").load("../bower_components/plyr/dist/sprite.svg");
```

js/everything.js
```diff
@@ -84,6 +85,33 @@ function onPlayerReady(event) {
...
+        plyrPlayer.source({
+            type: "video",
+            title: player.getVideoData().title,
+            sources: [{
+                src: currentVideoID,
+                type: "youtube"
+            }]
+        });
+        //Hide video; leave only controls
+        $(".plyr__video-wrapper").hide();
```
shakeelmohamed: This should be done with CSS

```css
.plyr__video-wrapper {
    display: none;
}
```

js/everything.js
```diff
@@ -84,6 +85,33 @@ function onPlayerReady(event) {
...
         ZenPlayer.init(currentVideoID);
+        setupPlyr();
+    }
+}
+
+function setupPlyr() {
+    //set up Plyr player
+    plyrPlayer = plyr.setup({
+        autoplay: true,
+        controls:["play", "current-time", "duration", "mute", "volume", "captions"]
```
shakeelmohamed: It doesn't show up anyways, but we don't need captions

shakeelmohamed:
First of all, thank you very much @BlinkBP for taking the initiative on this task! 

I played with this locally, some high level thoughts about bringing in plyr:

* Plyr is just using the same YouTube API we are under the hood
* By that logic, we should be able to remove our direct dependency on `https://www.youtube.com/iframe_api`
* But, Plyr doesn't appear to have a way for us to access the `YT.player` object, so we won't be able to access metadata (author, title) through Plyr (unless we send them some changes)
* We can't use Plyr without using the iframe API
* In order to get author/title, we need to fully load the video in *both* the Plyr and iframe players
* We can't use autoplay in Plyr because only 1 of the 2 players can play at a time - and the iframe one will load first (might be able to hack this a bit to work, but it's not worth the effort IMO)

From my current perspective, I see the following options for us (ordered from most to least favorable):

0. Send enough changes to Plyr via PR and make it work for us without breaking anything (I can reach out to the maintainer if we go this route)
0. Fork Plyr to the org, and heavily modify it to fit our needs
0. Find an alternative to Plyr, in #136 I mentioned audio.js, there may be others
0. Make additional ajax requests to the YouTube data API (like we do for description) for metadata (author, title, etc.)
0. Write/design our own player

BlinkBP:
@shakeelmohamed I'm happy to be here. :smiley:   
  
I think it's a good idea to change Plyr by PR to suit us more. This way we could have Plyr get YT metadata for us. After that we could incorporate Soundcloud and YT playlists support into Plyr (in the near future).  
Slapping together a simple media player seems like taking the easy way but if we make it "ours" by improving it (also we should mess around in plyr's css, just for our app) it should be nice.

js/everything.js
BlinkBP: There are and there will be issues with this while running the site locally (in `file` protocol). JS won't ever allow to access `file` protocol (websites would be able to steal everyone's data). `.load()` uses the same `XMLHttpRequest` but it's just one line instead of four.
I use python's `http.server` to quickly and easily host the site locally. Of course Youtube API has some issues when doing so.

index.html
```diff
@@ -8,25 +8,29 @@
...
     <script src="bower_components/seiyria-bootstrap-slider/dist/bootstrap-slider.min.js"></script>
     <script src="bower_components/typeahead.js/dist/typeahead.jquery.min.js"></script>
+    <script src="bower_components/plyr/dist/plyr.js"></script>
     <script src="js/zap-common.js"></script>
     <script src="js/everything.js"></script>
     <script src="https://www.youtube.com/iframe_api"></script>
+
 </head>
 <body>
+    <div id="plyr-svg" hidden></div>
```
shakeelmohamed: Let's move this down above `.plyr`, so they are logically grouped

index.html
```diff
@@ -87,6 +91,8 @@
             <br>
         </div>
         <br>
+        <div class="plyr"><audio></audio></div>
```
shakeelmohamed: Does the `<audio>` tag have any effect here? I didn't notice any difference when I was playing with it locally

js/everything.js
```diff
@@ -84,6 +85,28 @@ function onPlayerReady(event) {
     // Setup player
     if (currentVideoID) {
         ZenPlayer.init(currentVideoID);
+        setupPlyr();
+    }
+}
+
+function setupPlyr() {
+    //set up Plyr player
```
shakeelmohamed: We should return here if `plyrPlayer` is truthy. Not sure what would happen if one manually invoked `setupPlyr()` several times on the console :smile:

js/everything.js
```diff
@@ -84,6 +85,28 @@ function onPlayerReady(event) {
...
+    plyrPlayer = plyr.setup({
+        autoplay: true,
+        controls:["play", "current-time", "duration", "mute", "volume", "captions"]
+    })[0];
+    //Load video into Plyr player
+    if (plyrPlayer) {
+        plyrPlayer.source({
+            type: "video",
+            title: player.getVideoData().title,
```
shakeelmohamed: Is this title used anywhere by plyr?

index.html
BlinkBP: It doesn't work locally for me though.  
I'm pretty sure Plyr checks what's inside the div we're running Plyr on while initializing. It checks for `<div>`, `<audio>`, and `<video>` (used for YT and Vimeo, audio files, and video files respectively). When calling `.source` Plyr takes care of pushing new html elements but but before that, with empty div, it doesn't want to initialize properly for me.

shakeelmohamed: Which browser?

BlinkBP: Chrome. Doesn't work both locally and on local server.  
Also, `<div>` doesn't work either (js throws error). I think when seeing `<div>` Plyr looks for video id and such and can't find it. With `<audio>`, and `<video>` it doesn't care if there is `<source>` inside. Seems like a bug in Plyr for me.

js/everything.js
BlinkBP: Supposedly in Aria labeling.

index.html
shakeelmohamed: That's odd, I didn't see that behavior. We can revisit it later

BlinkBP: Should I revert? Reset? Plyr loads and is able to play with `<div>` but it stops somewhere with:

> Uncaught TypeError: Cannot read property 'play' of undefined

and because it stops there it doesn't even load controls' icons.

shakeelmohamed: Don't worry about it right now :smile:

shakeelmohamed:
Thanks to @SamPotts, if we upgrade to v1.5.17 we can do `element.plyr.embed.api('play')` to get every place where we have `player` currently. (Just checked and it's been updated on bower, you can just run `bower update plyr`)

Holler if you need help with the refactor

SamPotts:
Sorry if I was unclear, basically the `embed` property is a direct mapping to the respective api's. So all the methods for YouTube, such as `.getVideoData()` are available on that property. Vimeo uses `.api('method')` and is exposed in a similar way. 

So if you're using YouTube, an example would be:
```javascript
var data  = player.plyr.embed.getVideoData();
```
Don't forget a lot of these methods are available in the standardised API methods built into Plyr.

shakeelmohamed:
Thanks for the correction!

js/everything.js
```diff
@@ -84,6 +85,33 @@ function onPlayerReady(event) {
     // Setup player
     if (currentVideoID) {
         ZenPlayer.init(currentVideoID);
+        if (plyrPlayer) {
+            return;
+        }
+        else {
```
shakeelmohamed: nit: the `else` is redundant here since we return in the `if` block

BlinkBP:
@shakeelmohamed So what do we do now? Should I try to replace actual player with Plyr?

shakeelmohamed:
@BlinkBP exactly, you can start removing piece by piece. If you get stuck, I can make a PR against your branch to help move things along

index.html
BlinkBP: I spent about an hour trying to figure out why Plyr was throwing `Uncaught TypeError`. Finally I remembered that I didn't change those `<div>`'s to `<audio>` or `<video>`. I don't know if I should feel bad because I spent so much time figuring this out or good because I'm sure `<div>`'s don't work properly in there in our case. It doesn't matter that much but Plyr was throwing this error because `buttons` elements' references weren't there. It was trying to access `plyr.buttons.play` but `buttons` was undefined (Of course I'm talking about Plyr's code, not ours).

shakeelmohamed: I don't understand :smiley:

BlinkBP: > Don't worry about it right now :smile:  

So I forgot about it. And the same bug I had before happened again. I spent about an hour trying to figure it out before I remembered about this. More of a rant than something we should discuss. :stuck_out_tongue_winking_eye:

BlinkBP:
I pushed some changes but I'm stuck on two things:
Plyr provides `embed` with our precious `getVideoData()` much later, e.g. not right after we call `.source(...)`. With that we can't use `getVideoData()` right away to update Twitter message (`updateTweetMessage()` function).
Other thing, Plyr doesn't even consider autoplaying YT videos.
But if we could autoplay videos (further changes/features for Plyr) then we could listen for "play" event and with that event we would be sure that we can use `getVideoData()`. That's the only solution that comes to me at 4:40AM so there may be a better one. :wink:

shakeelmohamed:
@BlinkBP for the autoplay, I've made a PR: https://github.com/Selz/plyr/pull/181

Let me look into the rest of it

SamPotts:
I'll trigger an event on embed ready for you guys - should make life easier?

shakeelmohamed:
@SamPotts you might as well join the org since you're helping us out so much :grinning:

SamPotts:
Heh, well the issues you guys are facing are probably what everyone else is coming up against. It's no worries at all :+1:

SamPotts:
Added 'ready' in v1.5.18:
https://github.com/Selz/plyr#events

Let me know if you have any other annoyances and I will take a look :+1:

shakeelmohamed:
@SamPotts Thanks! I know there's some example code in the README for using events, but I haven't been able to get it to work. I must be missing something very obvious (using the latest Chrome on Mac)

The example code is:

```javascript
document.querySelector(".js-plyr").addEventListener("ready", function() {
  /* Magic happens */
});
```

So far, we have

```javascript
plyrPlayer = plyr.setup({
    autoplay: true,
    controls:["play", "current-time", "duration", "mute", "volume"]
})[0];
```

What I've tried:
* using the example code with the correct selector
  * noop
* giving `plyr.setup()` a selector as the first argument, then the options
  * noop
* replacing `document.querySelector('.js-plyr')` in the example code with `plyrPlayer` since that appears to be valid from the README (`var player = document.querySelector('.js-plyr');` and `var player = plyr.setup('.js-plyr')[0];` appear to be equivalent statements)
  * I end up getting `plyrPlayer.addEventListener` is undefined, even if I manually enter the code in the console after a video is fully loaded and/or playing

SamPotts:
The problem is that the value returned by plyr.setup() is the plyr *instance* rather than the *element* that the events are triggered on. Perhaps something like this:

```javascript
var player = document.querySelector(".js-player");

plyr.setup(player, {
    autoplay: true,
    controls:["play", "current-time", "duration", "mute", "volume"]
});

player.addEventListener("ready", function(event) { 
    // this is the plyr instance
});
```

If you need to access the plyr instance in that example above it'd be possible through `player.plyr` and then the `media` and `embed` are properties of that.

BlinkBP:
I don't really know why autoplay isn't working. And it seems that there is no `embed` property until a video starts playing.

SamPotts:
Ah damn :( - I'll check tonight...

SamPotts:
OK, so v1.5.20 should have fixes for both autoplay and accessing the `embed` property after the `ready` event has fired. I got the syntax a little wrong but have added notes here: 
https://github.com/Selz/plyr#embeds

:+1:

shakeelmohamed:
@SamPotts excellent, I'll take a look this week

shakeelmohamed:
@SamPotts the update + docs look great, I was able to hack together something in a few minutes.

@BlinkBP with this patch you should be able to finish up after you run `bower updater plyr` once more

```diff
diff --git a/js/everything.js b/js/everything.js
index 3afeac1..3c69099 100644
--- a/js/everything.js
+++ b/js/everything.js
@@ -1,14 +1,14 @@
 /*global getParameterByName, getSearchResults, getAutocompleteSuggestions, parseYoutubeVideoID, getYouTubeVideoDescription*/
 
-var player;
 var currentVideoID;
+var plyrPlayer;
 
 function initPlayer() {
     var currentVideoID = getCurrentVideoID();
 
     // Setup player
     if (currentVideoID) {
-        if (player) {
+        if (plyrPlayer) {
             return;
         }
         setupPlyr();
@@ -16,22 +16,22 @@ function initPlayer() {
 }
 
 function setupPlyr() {
-    player = document.querySelector(".plyr");
-
-    plyr.setup(player, {
+    var playerDiv = document.querySelector(".plyr");
+    plyrPlayer = plyr.setup(playerDiv, {
         autoplay: true,
         controls:["play", "current-time", "duration", "mute", "volume"]
-    });
+    })[0];
+
     //Inject svg with controls' icons
     $("#plyr-svg").load("../bower_components/plyr/dist/sprite.svg");
 
-    player.addEventListener("ready", function(event) {
+    playerDiv.addEventListener("ready", function(event) {
         onEmbedReady();
     });
 
     //Load video into Plyr player
-    if (player.plyr) {
-        player.plyr.source({
+    if (playerDiv.plyr) {
+        playerDiv.plyr.source({
             type: "video",
             title: "Title",
             sources: [{
@@ -63,7 +63,7 @@ function updateTweetMessage() {
     if (id) {
         opts.url += "/?v=" + id;
         //console.log(Object.keys(player.plyr));
-        opts.text = "I'm listening to " + player.plyr.embed.getVideoData().title;
+        opts.text = "I'm listening to " + plyrPlayer.plyr.embed.getVideoData().title;
     }
 
     twttr.widgets.createHashtagButton(
```

shakeelmohamed:
@BlinkBP how's this going :question:

BlinkBP:
It is working but there is still some stuff to do (variables/functions names, tests, and such). I'm not sure what to do with travis though.

shakeelmohamed:
@BlinkBP are the tests running locally for you? Not sure why the travis runs aren't being triggered, I can figure that out when we're done. Once you feel done, let me know and I can finish up and deal with the merge conflicts

shakeelmohamed:
@BlinkBP update on Travis, you can use this build for c6dc0f8 - 
https://travis-ci.org/zen-audio-player/zen-audio-player.github.io/jobs/119733843

I created #171 as a temp pull request space, I can manually update that branch until everything is done.

BlinkBP:
At the moment I think that's all there was to move from ZenPlayer to Plyr.
@shakeelmohamed Help with the travis and we'll be golden. :smiley:

shakeelmohamed:
@BlinkBP Great thanks! I've updated #171. I will manually resolve the merge conflicts created by 39cfdd373e6c4a2700e0b1c518b32d0ee26b53ce, I actually want to retain the structure of the ZenPlayer "class". The final merged code will look significantly different that what you've done. This is fine, just wanted to make sure you're aware of the remaining work :+1:

BlinkBP:
@shakeelmohamed Seems like I'm leaving you a lot of work. Good luck!
I've learned a lot throughout this PR. It's been a blast. :stuck_out_tongue_winking_eye:

shakeelmohamed:
@BlinkBP no worries, I've had worse :sweat_smile:

shakeelmohamed:
Closing in favor of #176

PR #157, opened 2016-03-12
Persist volume level to localStorage

### LINKED ISSUE #156: Persist volume via html5 localStorage
Problem: the volume seems to max out when opening a new tab, which is very bad for headphone users :smiley: 

The change is very similar to #70, but we actually do want to use `localStorage`, not `sessionStorage`

gdaolewe:
Closes #156.
The localStorage writes are debounced to avoid excessive disk writes.

### DIFF
--- a/js/everything.js
+++ b/js/everything.js
@@ -134,6 +134,12 @@ var ZenPlayer = {
         // Start video from where we left off
         player.seekTo(loadTime());
 
+        // Set volume to previous saved value, if any
+        var volume = loadVolume();
+        if (!isNaN(volume)) {
+            player.setVolume(volume);
+        }
+
         // Google Analytics
         ga("send", "event", "Playing YouTube video title", this.videoTitle);
         ga("send", "event", "Playing YouTube video author", this.videoAuthor);
@@ -228,7 +234,9 @@ var ZenPlayer = {
 
         function updateVolumeFromSlider() {
             if (player) {
-                player.setVolume($("#volume").slider("getValue"));
+                var volume = $("#volume").slider("getValue");
+                player.setVolume(volume);
+                storeVolume(volume);
             }
         }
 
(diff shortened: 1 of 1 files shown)


js/everything.js
```diff
@@ -402,6 +410,28 @@ function loadTime() {
...
+    if (debouncedPersistVolumeTimer) {
+        clearTimeout(debouncedPersistVolumeTimer);
+    }
+    debouncedPersistVolumeTimer = setTimeout(function () {
+        localStorage.setItem(LOCAL_STORAGE_VOLUME_KEY, volume);
+    }, VOLUME_PERSIST_DELAY);
+}
+
+function loadVolume() {
+    return parseInt(localStorage.getItem(LOCAL_STORAGE_VOLUME_KEY));
```
shakeelmohamed: `parseInt()` should have the second param (radix) set to `10` - the default is 8 and can lead to some oddities. [Good Stack Overflow explanation](http://stackoverflow.com/questions/850341/how-do-i-work-around-javascripts-parseint-octal-behavior)

js/everything.js
```diff
@@ -402,6 +410,28 @@ function loadTime() {
...
+var LOCAL_STORAGE_VOLUME_KEY = "VOLUME_LEVEL";
+
+// The delay to debounce writes of the volume level to localStorage
+var VOLUME_PERSIST_DELAY = 200;
+
+// The timeout ID tracking debounced writes of volume level to localStorage
+var debouncedPersistVolumeTimer;
+
+function storeVolume(volume) {
+    if (debouncedPersistVolumeTimer) {
```
shakeelmohamed: How important do you think it is to limit writing to `localStorage` by time? I'm not very familiar with the API

In my head I was thinking of doing something simpler, (pseudocode):

```javascript
if (currentVolume !== localStorage.getItem("volume")) {
    localStorage.setItem("volume", currentVolume);
}
```

js/everything.js
```diff
@@ -402,6 +410,28 @@ function loadTime() {
     }
 }
 
+// Key to the persisted volume level entry in localStorage
+var LOCAL_STORAGE_VOLUME_KEY = "VOLUME_LEVEL";
+
+// The delay to debounce writes of the volume level to localStorage
+var VOLUME_PERSIST_DELAY = 200;
```
shakeelmohamed: Is this too slow, too fast? I'm curious how you arrived at this number

shakeelmohamed:
Thank you for doing this so quickly @gdaolewe! Let me know if my comments are unclear

shakeelmohamed:
@gdaolewe I just realized plyr (coming in with #153) already used localStorage for persisting the volume levels. If you still want to do this work, know that it will go away once we switch over to plyr. Sorry about that :blush:

js/everything.js
gdaolewe: Ah yes, a fact I once knew and forgot.

js/everything.js
gdaolewe: It would be up to you (though moot given the comments below), but FYI `localStorage.setItem()` is a blocking disk write. The size of the write here is tiny, so it's unlikely to result in noticeable UI stutter in practice; I think the more relevant concern for an application like this would be minimizing energy consumption for laptop users.

js/everything.js
gdaolewe: I meant to bring that up as a matter for discussion, I don't really know what an appropriate number would be. The tradeoff is between potentially doing a lot of disk writes if a user adjusts the volume every few seconds, and debouncing by a much larger delay and potentially losing the value if the user navigates away or their browser crashes. 200ms is just in the range of numbers we use for this kind of thing at my work, but usually the data being persisted there is more critical.

gdaolewe:
> I just realized plyr (coming in with #153) already uses localStorage for persisting the volume levels. If you still want to do this work, know that it will go away once we switch over to plyr. Sorry about that 

No worries, I'll close this in that case.

Issue #167, opened 2016-03-16
Update screenshot


shakeelmohamed:
Thanks to @YuriBrunetto we recently got an awesome redesign! We need an updated screenshot for the README :tada:

YuriBrunetto:
What if we use a mockup? It'll shows both web and mobile screens

shakeelmohamed:
Why not real screenshots?

YuriBrunetto:
What I mean was: using **real** screenshots in a mockup. :smile:

shakeelmohamed:
Go for it

PR #168, opened 2016-03-17
Update screenshot #167

YuriBrunetto:
@shakeelmohamed what do you think?


shakeelmohamed:
I like the gradient design in the background, but I think showing a phone is a bit misleading since it doesn't really work on iOS.

YuriBrunetto:
@shakeelmohamed hmm.. so, do I remove the iPhone?

shakeelmohamed:
@YuriBrunetto I think so. We might consider making an iOS app, but I don't know if that's even a reasonable project since you can play YouTube videos while using other apps

YuriBrunetto:
@shakeelmohamed what about now? :smile:

shakeelmohamed:
I like it! Before I merge it, do you have a template or project file for the image? If so, I can make a new repo called "design resources" and you can check it in there. The goal is to make updating the screenshot possible for anyone

YuriBrunetto:
Yes, I have the .psd here!

shakeelmohamed:
@YuriBrunetto see https://github.com/zen-audio-player/design-resources/issues/1

PR #190, opened 2016-04-26
Trim down content in CONTRIBUTING.md, move detailed content to Wiki.

### LINKED ISSUE #49: Docs
We basically have 0 docs right now, a wiki would be nice :smile: 
Maybe we even make a separate repo just for docs? Then we can have zen-audio-player.github.io/docs be a valid URL

josectello:
Moved **Syncing** and other content to Contributions page of Wiki. Added a screenshot for submitting a PR (banner that appears on GitHub). 

Potentially a few more bullets that I think may be able to added into `CONTRIBUTING.md` without bloating the file too much.

Some ideas for the sections in the Contributions page of the Wiki:
* Squashing commits
* Maintaining a local fork of the repo (over time)

This (potentially) closes #49. But probably not, since this still needs feedback!


.github/CONTRIBUTING.md
```diff
@@ -17,28 +15,9 @@ Join the chat room on Gitter
 0. Make your changes.
 0. Run the tests: `npm test`.
 0. Run `git commit` and add your message describing what you have changed. If you have made changes that address an issue include the text `Closes #1` (where 1 would be the issue number) to your commit message.
-0. Submit a [pull request](https://github.com/zen-audio-player/zen-audio-player.github.io/pulls) - Travis CI will run tests to make sure nothing broke.
-0. In case the following occurs, skip to **[Syncing](#syncing)** below  
+0. Submit a [pull request](https://github.com/zen-audio-player/zen-audio-player.github.io/pulls) - Travis CI will run tests to make sure nothing broke. Once you have committed your changes locally and pushed them to your fork of the repo, a banner with a **Compare & pull request** button should appear on the main Zen Audio Player repo page to submit a pull request.
+![submit pr](/.github/submit_pr.png)
```
shakeelmohamed: This may be a broken link, I don't think you need the leading `/` either way

shakeelmohamed:
Thanks @josectello, we should do the same for the PR & issue templates. I fear they're too long to actually be used.

For the wiki, can you create another page and move your additions there? The ["Contribution Flow"](https://github.com/zen-audio-player/zen-audio-player.github.io/wiki/Contribution-flow) page was meant to only be that first section I added.

The actual changes to `CONTRIBUTING.md` look fine. After you add the new wiki page, you should be able to link to a specific heading within a wiki page like you do in the steps.

josectello:
@shakeelmohamed I'll get on that today. Thanks!

.github/CONTRIBUTING.md
josectello: Are you referring to the pull request link? It's working for me.
And are you saying `github.com/zen-audio-player/zen-audio-player.github.io/pulls` is preferable over `https://github.com/zen-audio-player/zen-audio-player.github.io/pulls`?

josectello:
@shakeelmohamed You're probably very busy as well, but I was wondering if you saw my line comment about link above.

.github/CONTRIBUTING.md
shakeelmohamed: I think `/.github/submit_pr.png` might not need the first `/`, not sure if it will work with it there

josectello: Ah! I was looking at the wrong link, that makes sense.

josectello: Now that I took a closer look, the link for that other screenshot: `/.github/outofsync.png` has the same leading backslash and seems to link properly. Would you still recommend removing them (from both links)?

shakeelmohamed: Yeah, when I see a path beginning with `/` I think of the root of a *nix filesystem... which is probably not what we want here even if it's valid 😆

shakeelmohamed:
Looks great, thanks @josectello! For the record, leave a comment after you've made changes otherwise I won't get a notification 😴

Issue #192, opened 2016-05-01
Resolve CSS hacks using !important


shakeelmohamed:
See the list of CSS warnings due to the use of `!important`: https://codeclimate.com/github/zen-audio-player/zen-audio-player.github.io/pull/176

Done criteria:

- [ ] Achieve the same UI effect without using `!important` in `css/styles.css`
- [ ] remove HACK comment in the stylesheet
- [ ] add the stylelint rule back to `.stylelintrc`: `"declaration-no-important": true,`

yoshrubin:
I'd love to work on this, I would need a bit of direction regarding what exactly needs to be fixed @shakeelmohamed

shakeelmohamed:
@yoshrubin great, ping me on our gitter chat room if this isn't clear.

At the bottom of `css/styles.css` I've temporarily added some CSS hacks to workaround Plyr's new UI. The goal is to achieve this same effect without using `!important` in the CSS rules.

```css
/* HACK: override the styles from ".plyr__audio .plyr__controls" */
.plyr__controls {
    padding: 10px !important;
    border-radius: inherit !important;
    background: #ffffff !important;
    border: 1px solid #dbe3e8 !important;
    box-shadow: 0 1px 1px rgba(0, 0, 0, 0.05) !important;
    color: #565d64 !important;
    position: relative !important;
}
```

To understand the visual effects, here's the before & after

![](https://cloud.githubusercontent.com/assets/1480063/14940126/ff2cf896-0f1b-11e6-8c12-07bed62acda8.png)
![](https://cloud.githubusercontent.com/assets/1480063/14940200/91b03828-0f20-11e6-85f2-be42a7db890b.png)

PR #195, opened 2016-05-02
#192 fix

yoshrubin:
fixed css !important hack


yoshrubin:
found errors when running npm test

Issue #202, opened 2016-05-11
Update the demo logic to use a random video every time


shakeelmohamed:
We currently have a single hardcoded song, and it's an amazing song! 😀 
I'm thinking we can add a bit of variety now by (for now) storing a hardcoded array of songs by video ID.

Some song suggestions:

- [Cazzette - Together (Lost Kings Remix)](https://www.youtube.com/watch?v=EBerFisqduk)
- [Disclosure - Magnets (A-Trak Remix) ft. Lorde](https://www.youtube.com/watch?v=zBIE8waeHCI)
- [The Temper Trap - Sweet Disposition](https://www.youtube.com/watch?v=4C8e7nNLZNs)

Requirements;

- [ ] Add an array of video IDs of at least 3 demo songs, including the one we already have
- [ ] Add a function to return a random value from that array
- [ ] Update the `$("#demo").hide();` logic to hide the demo button for any `currentVideoID` in that array
- [ ] Update the `$("#demo").click()` event handler to use the new function

Bonus (@shakeelmohamed to do this otherwise):

- [ ] Add a test to verify that all demo video IDs are valid

shakeelmohamed:
Done in #203 by @hjh17!

Issue #20, opened 2015-05-18
Fix bug in demo link continuing concatenation


shakeelmohamed:
If the demo link fails, meaning there's a problem with JS/YouTube's API, the URL ends up looking like http://zen-audio-player.github.io/?v=koJv-j1usoI?v=koJv-j1usoI or something really strange. Just need to do a simple indexOf() check somewhere - or improve error handling.. maybe needs a giant try/catch somewhere

PR #203, opened 2016-05-13
randomized demo

hjh17:
Hey whatsup? It's my first contribution to open source so as much feedback as possible would be much appreciated.

I was looking for open source contribution and I came across this [https://github.com/zen-audio-player/zen-audio-player.github.io/issues/202]. I figured that this would be easy to implement. However, I could not go through the unit testing part since I'm not familiar with unit testing in general. I am interested in adding the test though if someone could help me get started.

### Description
- On click demo, plays a randomized song from a given array.

### DIFF
--- a/js/everything.js
+++ b/js/everything.js
@@ -518,8 +518,13 @@ $(function() {
     });
 
     // Some demo video's audio, feel free to add more
-    var demos = ["koJv-j1usoI", "5cJIvC6AAkc" , "EBerFisqduk", "DlKXJ906pd8"];
-    function playDemo(){
+    var demos = [
+        "koJv-j1usoI", // The Glitch Mob - Starve the Ego, Feed the Soul
+        "5cJIvC6AAkc", // Family Force 5 - Dance Or Die Official Music Video
+        "EBerFisqduk", // Cazzette - Together (Lost Kings Remix)
+        "DlKXJ906pd8" // Ronald Jenkees - Throwing Fire
+    ];
+    function pickDemo() {
         return demos[Math.floor(Math.random() * demos.length)];
     }
 
@@ -527,7 +532,6 @@ $(function() {
     if ($.inArray(currentVideoID, demos) !== -1) {
         $("#demo").hide();
     }
-
     // Handle demo link click
     $("#demo").click(function(event) {
         event.preventDefault();
@@ -536,9 +540,9 @@ $(function() {
 
         // Don't continue appending to the URL if it appears "good enough".
         // This is likely only a problem if the demo link didn't work right the first time
-        if (window.location.href.indexOf(playDemo()) === -1) {
-            window.location.href = makeListenURL(playDemo());
-
+        var demoPicked = pickDemo();
+        if (window.location.href.indexOf(demos) === -1) {
+            window.location.href = makeListenURL(demoPicked);
         }
         else {
             ga("send", "event", "demo", "already had video ID in URL");
--- a/js/everything.js
+++ b/js/everything.js
@@ -414,6 +414,18 @@ function wrapParseYouTubeVideoID(url) {
     }
 }
 
+// Some demo video's audio, feel free to add more
+var demos = [
+    "koJv-j1usoI", // The Glitch Mob - Starve the Ego, Feed the Soul
+    "5cJIvC6AAkc", // Family Force 5 - Dance Or Die Official Music Video
+    "EBerFisqduk", // Cazzette - Together (Lost Kings Remix)
+    "DlKXJ906pd8" // Ronald Jenkees - Throwing Fire
+];
+
+function pickDemo() {
+    return demos[Math.floor(Math.random() * demos.length)];
+}
+
 $(function() {
     // Keen.io
     client = new Keen({ //eslint-disable-line no-undef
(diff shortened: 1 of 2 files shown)


js/everything.js
```diff
@@ -517,10 +517,14 @@ $(function() {
         }
     });
 
-    var starveTheEgoFeedTheSoulGlitchMob = "koJv-j1usoI";
+    // Some demo video's audio, feel free to add more
+    var demos = ["koJv-j1usoI", "5cJIvC6AAkc" , "EBerFisqduk", "DlKXJ906pd8"];
```
shakeelmohamed: Let's make these 1 per line with a comment saying the title for each, ie:

```javascript
var demos [
  "foo", // Starve the ego, feed the soul - the glitch mob
  "bar" // Cool song - random artist
];
```

js/everything.js
```diff
@@ -517,10 +517,14 @@ $(function() {
...
+    function playDemo(){
```
shakeelmohamed: nit: rename to something like `pickDemo()`

js/everything.js
```diff
@@ -532,8 +536,9 @@ $(function() {
 
         // Don't continue appending to the URL if it appears "good enough".
         // This is likely only a problem if the demo link didn't work right the first time
-        if (window.location.href.indexOf(starveTheEgoFeedTheSoulGlitchMob) === -1) {
-            window.location.href = makeListenURL(starveTheEgoFeedTheSoulGlitchMob);
+        if (window.location.href.indexOf(playDemo()) === -1) {
+            window.location.href = makeListenURL(playDemo());
```
shakeelmohamed: This is a bug. We should call `playDemo()`, and store it's value before this if statement because we'll get 2 different values back from `playDemo()`.

That aside, I think the if condition may need to change to something like (pseudocode)

```javascript
`if (window.location.href does not contain any element from the demo array) {
    // makeListenURL() ....
}
```

shakeelmohamed:
Thanks so much @hjh17! Welcome aboard, this is a great start so far. I've left you come comments inline let me know if anything is unclear.

The test is failing because the demo functionality won't just play the previously hardcoded demo only.
The quick solution would be to hardcode the demos array in `test/index.js`, the [test code](https://github.com/zen-audio-player/zen-audio-player.github.io/blob/master/test/index.js#L132) is pretty simple.

To run tests locally, you need Node.js 4.x or later installed. Then `npm install` at the root of this project, then `npm test`. The tests will also do some code quality checks (eslint for JavaScript style, etc.) before running the tests

hjh17:
Thank you for the nice feedback @shakeelmohamed I feel like I'm learning a lot ! I fixed all the mistakes and made adjustments towards the overall coding style.

I came across the unit test problem when I made the pull request but didn't have enough confidence to tackle it. I dived straight into that problem after receiving your feedback and I might have a working solution. Let me know what you think.

js/everything.js
```diff
@@ -518,16 +518,20 @@ $(function() {
...
     // Some demo video's audio, feel free to add more
-    var demos = ["koJv-j1usoI", "5cJIvC6AAkc" , "EBerFisqduk", "DlKXJ906pd8"];
-    function playDemo(){
+    var demos = [
+        "koJv-j1usoI", // The Glitch Mob - Starve the Ego, Feed the Soul
+        "5cJIvC6AAkc", // Family Force 5 - Dance Or Die Official Music Video
+        "EBerFisqduk", // Cazzette - Together (Lost Kings Remix)
+        "DlKXJ906pd8" // Ronald Jenkees - Throwing Fire
+    ];
+    function pickDemo() {
```
shakeelmohamed: Can we move this function outside this block of code? It should be above `$(function() {`

js/everything.js
```diff
@@ -536,9 +540,9 @@ $(function() {
 
         // Don't continue appending to the URL if it appears "good enough".
         // This is likely only a problem if the demo link didn't work right the first time
-        if (window.location.href.indexOf(playDemo()) === -1) {
-            window.location.href = makeListenURL(playDemo());
-
+        var demoPicked = pickDemo();
```
shakeelmohamed: This looks fine.

Not sure if this is even an achievable scenario, if it is we should make sure none of the demo IDs exist in the URL

* url is `zenplayer.audio?v=demo1`
* we get to this block of code trying to add a new random demo
* url becomes `zenplayer.audio?v=demo1&v=demo1`

js/everything.js
```diff
@@ -532,8 +540,9 @@ $(function() {
 
         // Don't continue appending to the URL if it appears "good enough".
         // This is likely only a problem if the demo link didn't work right the first time
-        if (window.location.href.indexOf(starveTheEgoFeedTheSoulGlitchMob) === -1) {
-            window.location.href = makeListenURL(starveTheEgoFeedTheSoulGlitchMob);
+        var demoPicked = pickDemo();
```
shakeelmohamed: nit: rename to pickedDemo

shakeelmohamed:
@hjh17 You're welcome, the changes look pretty good! Left a comment on some minor things, test code looks good though

hjh17:
I'm not quite sure what you mean by the double append part. The only weird scenario I see is the one when a user happens to input one of the demo songs then wanting to hit the demo button.

shakeelmohamed:
@hjh17 see #20

hjh17:
@shakeelmohamed I'm not finding any way to emulate this problem. I tried breaking the APIkey and adding no resulting demos, still I never get that double append url...

shakeelmohamed:
@hjh17 ok, when we added Plyr support that bug might be possible anymore. We can always make a bugfix if we see it again 😄  Thanks for your efforts, welcome to the org!

hjh17:
Amazing experience ! Thank you for the guidance @shakeelmohamed . Feel free to tag me if there is any noob work to be done.

PR #206, opened 2016-05-20
Closes #193 Add support for linking to a specific time position

### LINKED ISSUE #193: Add support for linking to a specific time position
In YouTube videos, you can set the `t` query string parameter to the number of seconds you want to start the video at. IE: [`https://www.youtube.com/watch?v=koJv-j1usoI&t=30`](https://www.youtube.com/watch?v=koJv-j1usoI&t=30) to start 30 seconds into the video.

General guidance:

* Call `getParameterByName(window.location.href, "t")` to get the value from the query parameter
* Update `window.sessionStorage[videoID]` in `ZenPlayer.init()` with that value, and it should "just work" due to the event listener for `playing`

Done criteria:

- [ ] No change in functionality if the `t` parameter is missing
- [ ] If `t` parameter is present it immediately overrides the value in `window.sessionStorage[videoID]`, and plays without continually trying to seek.

Bonus:

- [ ] Figure out how to update the URL with the updated current time if the `t` parameter is present (without triggering a page reload). To pull this off, you'll likely need to use a hash URL scheme instead of query arguments (ie: `?v=foobar#t123` instead of `?v=foobar&t=123`

hjh17:
@shakeelmohamed check this out !

### Types of changes

### Description 
Now we are able to link to a specific time position both from the URL and the input bar.

I had some minor issues, i.e. if you put &t=90 on a 89 second video on youtube, it ends the video. I wanted to replicate that with something like

```Javascript
var t = getCurrentTimePosition();
if(t > plyrPlayer.plyr.embed.getDuration()){
        t = plyrPlayer.plyr.embed.getDuration();
}
```
but that just puts the song back on time 00:00. Even though I changed the criteria to -5 seconds before `getDuration()`, it still sent me to 00:00. 

But anyways, I think it's a good solution. There was no need for URL hash scheme approach for this one, I had no problems calling 'makeListenURL(videoID)' with '&t=XXX' as input.

### Final checklist:
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] All tests passed.


shakeelmohamed:
@hjh17 thanks! Can you update your branch? I've made some changes to the eslint configuration that will catch the minor style issues I'd normally comment on in-line.

js/everything.js
```diff
@@ -490,13 +508,17 @@ $(function() {
     // Handle form submission
     $("#form").submit(function(event) {
         event.preventDefault();
-
         var formValue = $.trim($("#v").val());
+        var formValueTime = formValue.match(/(&t=\d*)$/g);
```
shakeelmohamed: Can we use `parseInt()` instead?

js/everything.js
```diff
@@ -509,13 +531,17 @@ $(function() {
                         key: youTubeDataApiKey,
                         part: "snippet",
                         fields: "items/snippet/description",
-                        id: videoID
+                        id: videoID,
+                        startSeconds:50
```
shakeelmohamed: this should be replaced with a variable

js/everything.js
```diff
@@ -368,12 +380,11 @@ function removeSearchQueryFromURL(url) {
     }
     return url;
 }
-
 function makeListenURL(videoID) {
```
shakeelmohamed: This function should be updated to a take an optional second parameter for position/time, etc.

js/everything.js
hjh17: This gathers the whole string `&t=X` which I put straight into the `makeListenURL` function.

js/everything.js
hjh17: this `startSeconds:50` line should have not been there...

js/everything.js
hjh17: I updated the function with a second parameter

hjh17:
@shakeelmohamed I can check out the unit test for the parameter `t`tomorrow.

js/everything.js
shakeelmohamed: In that case, why not use `getParameterByName(formValue, "t");`?

js/everything.js
```diff
@@ -368,13 +380,16 @@ function removeSearchQueryFromURL(url) {
...
 }
-
-function makeListenURL(videoID) {
+function makeListenURL(videoID,videoPosition) {
     var url = removeSearchQueryFromURL(window.location.href);
     // Remove any #s which break functionality
     url = url.replace("#", "");
-
-    return url + "?v=" + videoID;
+    if (videoPosition) {
```
shakeelmohamed: Let's clean this up to have a single `return`. The `videoPosition` param should just be a `Number` not the full `&t=x` string

js/everything.js
```diff
@@ -438,8 +453,15 @@ $(function() {
...
     var currentVideoID = getCurrentVideoID();
+    var currentVideoPosition = getCurrentTimePosition();
+    if (currentVideoPosition) {
+        currentVideoPosition = "&t=" + currentVideoPosition;
+    }
+    else {
+        currentVideoPosition = "";
+    }
     if (currentVideoID) {
-        $("#v").attr("value", currentVideoID);
```
shakeelmohamed: Does this update the form to be something like `foo&t=123` where `foo` is the video ID? We don't want that. So, we can simplify all of this code

shakeelmohamed:
> @shakeelmohamed I can check out the unit test for the parameter `t` tomorrow.

@hjh17 have fun 😄

js/everything.js
hjh17: So basically we have two cases:

1. User inputs time position in the URL, `getParameterByName(formValue, "t");` retrieves that time position.

2. User inputs time position in the form, this section of the code deals with that.

js/everything.js
hjh17: You mean that the form should only contain the video ID in any case? So if user inputs i.e. `koJv-j1usoI&t=30` in the form, it should put the song at position 0:30 and display `koJv-j1usoI` ? I

hjh17:
@shakeelmohamed I am having a bit trouble making the test for `getCurrentTimePosition()` function. My idea is this:

* Copy paste the whole `getCurrentTimePosition()` function inside test/index.js
* Inside `describe("JavaScript components"... ` I will put the test
* Make a couple of tests that might look something like:
 * If `getCurrentTimePosition()` < `plyrPlayer.plyr.embed.getDuration();` then `t` should be somewhere in between zero and `plyrPlayer.plyr.embed.getDuration();`
 * If `getCurrentTimePosition()` > `plyrPlayer.plyr.embed.getDuration();` then `t=0`
 * If `getCurrentTimePosition()` = `plyrPlayer.plyr.embed.getDuration();` then `t=0`
 * If `getCurrentTimePosition()`  < `0` then `t=0`
 * If `getCurrentTimePosition()`  = `0` then `t=0`

My problem is that I'm not so familiar with the unit testing concept and it's environment. For example the scope: can I access every variable/function from `js/everything.js` with `var Browser = require("zombie");`?

js/everything.js
```diff
@@ -368,13 +380,17 @@ function removeSearchQueryFromURL(url) {
...
+function makeListenURL(videoID, videoPosition) {
     var url = removeSearchQueryFromURL(window.location.href);
     // Remove any #s which break functionality
     url = url.replace("#", "");
-
-    return url + "?v=" + videoID;
+    if (videoPosition) {
+        videoPosition = "&t=" + videoPosition;
+    }
+    else {
```
shakeelmohamed: This isn't how I would've written it 😄  
I was thinking more like...

```javascript
url += "?v= + videoID;
if (videoPosition) {
  url += "&t=" + videoPosition;
}
return url;
```

js/everything.js
shakeelmohamed: OK, should the regex be `/(&t=\d+)$/g` instead since we want 1 or more digits?

js/everything.js
```diff
@@ -490,13 +506,16 @@ $(function() {
     // Handle form submission
     $("#form").submit(function(event) {
         event.preventDefault();
-
         var formValue = $.trim($("#v").val());
+        var formValueTime = formValue.match(/(&t=\d*)$/g);
+        if (formValueTime) {
+            formValueTime = parseInt(formValueTime[0].substring(3));
```
shakeelmohamed: `parseInt()` always should have `10` as the second parameter for radix.

js/everything.js
```diff
@@ -490,13 +506,16 @@ $(function() {
     // Handle form submission
     $("#form").submit(function(event) {
         event.preventDefault();
-
         var formValue = $.trim($("#v").val());
+        var formValueTime = formValue.match(/(&t=\d*)$/g);
+        if (formValueTime) {
```
shakeelmohamed: We should also check for the length of `formValueTime`

js/everything.js
shakeelmohamed: Actually, we should change the regex to be `/&t=(\d+)$/` since we only want to capture the digits. We can remove the global search as well since we just want the first value

js/everything.js
```diff
@@ -357,6 +364,11 @@ function getCurrentVideoID() {
     return v;
 }
 
+function getCurrentTimePosition() {
+    var t = getParameterByName(window.location.search, "t");
```
shakeelmohamed: We should be trying to `parseInt(t, 10)` here, and only return the value if it's valid (ie: not `NaN`)

js/everything.js
hjh17: Lol... 😃

hjh17:
@shakeelmohamed Check me out !

js/everything.js
shakeelmohamed: Still some work remaining here

js/everything.js
```diff
@@ -357,6 +364,16 @@ function getCurrentVideoID() {
     return v;
 }
 
+function getCurrentTimePosition() {
+    var t = parseInt(getParameterByName(window.location.search, "t"), 10);
+    if (t >= 0 && t < 999999) {
```
shakeelmohamed: Should use `if (t > 0 && t < Number.MAX_VALUE)`, we can ignore it if the value is `0` since it's' noop.

js/everything.js
```diff
@@ -357,6 +364,16 @@ function getCurrentVideoID() {
...
+        return t;
+    }
+    else {
```
shakeelmohamed: We don't even need this `else` since the return value will be `undefined` anyways

js/everything.js
```diff
@@ -516,7 +538,7 @@ $(function() {
                             window.location.href = makeSearchURL(formValue);
                         }
                         else {
-                            window.location.href = makeListenURL(videoID);
+                            window.location.href = makeListenURL(videoID, formValueTime);
```
shakeelmohamed: If `formValueTime` is set to an empty array from the regex match above (ie: no matches) we'll pass an empty array to `makeListenURL()` which is evaluated to a truthy value. I believe this means the URL will end up being something like `https://zenplayer.audio/?v=foo&t=` if no `t` param was passed  in.

shakeelmohamed:
@hjh17 on the unit testing code, you should have access to everything if you do something like `browser.evaluate("window. getCurrentTimePosition()")`. You might run into timing issues, if so don't worry about the tests for now. You're welcome to spend time experimenting with Zombie, the [docs](http://zombie.js.org/) may be helpful.

hjh17:
@shakeelmohamed I fixed those things. I didn't know about the thing with functions returning undefined nor the `Number.MAX_VALUE`. Always learning new things 😄 . 

I changed the regexp from `/(&t=\d*)$/g` to `/(&t=\d+)$/g` so it should only match if there is at least 1 digit.

As for the unit test, I think I might need to do some more research on Unit testing in general before I tackle that stuff.

hjh17:
I also had an idea. When you are listening to a song in Zen and click on the Youtube link, you should be sent to the Youtube video at the same time position you currently were in Zen.

js/everything.js
```diff
@@ -490,13 +506,16 @@ $(function() {
     // Handle form submission
     $("#form").submit(function(event) {
         event.preventDefault();
-
         var formValue = $.trim($("#v").val());
+        var formValueTime = formValue.match(/(&t=\d+)$/g);
```
shakeelmohamed: Still need to update the regex to only capture the digits, ie: `/&t=(\d+)$/g`

shakeelmohamed:
> I also had an idea. When you are listening to a song in Zen and click on the Youtube link, you should be sent to the Youtube video at the same time position you currently were in Zen.

@hjh17 Of course, great idea! Outside the if statement in the `timeupdate` event listener, you can add a bit of jQuery to update the `href` attribute on `#zen-video-title`

hjh17:
Yo @shakeelmohamed, check me out. I changed a bit in the regexp. I updated the if statement to `>1` since the .exact returns the whole match and the match inside brackets. 

I also added deliverance of current position to the Youtube's url. Let me know what you think.

js/everything.js
```diff
@@ -203,10 +211,11 @@ var ZenPlayer = {
 
             plyrPlayer.addEventListener("timeupdate", function() {
                 // Store the current time of the video.
+                var resumeTime;
                 if (window.sessionStorage) {
                     var currentTime = plyrPlayer.plyr.embed.getCurrentTime();
                     var videoDuration = plyrPlayer.plyr.embed.getDuration();
-                    var resumeTime = 0;
+                    resumeTime = 0;
```
shakeelmohamed: why is this set to `0` instead of `currentTime`?

shakeelmohamed: move this initialization outside the `if`

js/everything.js
```diff
@@ -218,6 +227,8 @@ var ZenPlayer = {
                     }
                     window.sessionStorage[videoID] = resumeTime;
                 }
+                var updatedUrl = that.videoUrl + "&t=" + parseInt(resumeTime, 10);
```
shakeelmohamed: Why is `parseInt()` used here? Won't `resumeTime` always be a `Number`?

Also, these 2 lines should be surrounded in an `if (resumeTime > 0)` block - there's no point in updating the URL if we're just adding `&t=0`

js/everything.js
```diff
@@ -357,6 +368,13 @@ function getCurrentVideoID() {
     return v;
 }
 
+function getCurrentTimePosition() {
+    var t = parseInt(getParameterByName(window.location.search, "t"), 10);
+    if (t > 0 && t < Number.MAX_VALUE) {
+        return t;
+    }
```
shakeelmohamed: Thinking about this again, let's `return 0;` outside the `if`

js/everything.js
```diff
@@ -490,13 +510,16 @@ $(function() {
     // Handle form submission
     $("#form").submit(function(event) {
         event.preventDefault();
-
         var formValue = $.trim($("#v").val());
+        var formValueTime = /&t=(\d+)$/g.exec(formValue);
+        if (formValueTime && formValueTime.length > 1) {
+            formValueTime = parseInt(formValueTime[1], 10);
```
shakeelmohamed: It looks like we need 2 matches now, what's the value of `formValueTime` before the `if`?

js/everything.js
```diff
@@ -490,13 +510,16 @@ $(function() {
...
+            formValue = formValue.replace(/(&t=\d*)$/g, "");
```
shakeelmohamed: Why `\d*` here, but `\d+` above? Also, why are the parentheses in this regex? I don't think they're needed since we're not capturing a group

shakeelmohamed:
Sorry for the delay @hjh17, left some more comments inline

js/everything.js
shakeelmohamed: Actually, `if (rescueTime <= 0)` we should do `$("#zen-video-title").attr("href", that.videoUrl);`. To be effecient though, we should only do it `if ($("#zen-video-title").attr("href") !== that.videoUrl)`.

hjh17:
@shakeelmohamed thanks for the comments. I'm in middle of my exam period right now. I will definitely keep working on this after that !

shakeelmohamed:
ping @hjh17

hjh17:
thanks for reminding me @shakeelmohamed I will certainly work on this tomorrow ! 😄

hjh17:
@shakeelmohamed Check me out !

js/everything.js
hjh17: I replaced `parseInt()` with `Math.round()`, since youtube won't accept time positions with decimal places.

js/everything.js
hjh17: Not sure what you mean by needing 2 matches. `formValueTime` returns e.g `["&t=40", "40", index: 11, input: "EBerFisqduk&t=40"]`

js/everything.js
hjh17: The `\d+` was a mistake. I removed the second parentheses, I needed the first one to capture the time value.

js/everything.js
```diff
@@ -371,13 +394,15 @@ function removeSearchQueryFromURL(url) {
     }
     return url;
 }
-
```
shakeelmohamed: revert this line, there should be a blank line between functions

js/everything.js
```diff
@@ -221,6 +229,13 @@ var ZenPlayer = {
                     }
                     window.sessionStorage[videoID] = resumeTime;
                 }
+                if (resumeTime > 0) {
+                    var updatedUrl = that.videoUrl + "&t=" + Math.round(resumeTime);
+                    $("#zen-video-title").attr("href", updatedUrl);
+                }
+                else if (resumeTime <= 0 && $("#zen-video-title").attr("href") !== that.videoUrl) {
+                    $("#zen-video-title").attr("href", that.videoUrl);
+                }
```
shakeelmohamed: Let's clean this up so we only set the `href` attribute once, using the if/else to change the `updateUrl` value then set it afterwards

js/everything.js
```diff
@@ -229,13 +229,15 @@ var ZenPlayer = {
...
                 }
+                var updatedUrl;
                 if (resumeTime > 0) {
-                    var updatedUrl = that.videoUrl + "&t=" + Math.round(resumeTime);
+                    updatedUrl = that.videoUrl + "&t=" + Math.round(resumeTime);
                     $("#zen-video-title").attr("href", updatedUrl);
                 }
                 else if (resumeTime <= 0 && $("#zen-video-title").attr("href") !== that.videoUrl) {
-                    $("#zen-video-title").attr("href", that.videoUrl);
+                    updatedUrl = that.videoUrl;
```
hjh17: Not sure if we need to wrap `updatedUrl` in a `if (updatedUrl)` statement. Can't imagine any case where the variable will remain empty. What do you think?

hjh17:
@shakeelmohamed updated !

js/everything.js
shakeelmohamed: to be safe, initialize `updatedUrl` to `that.videoUrl`

hjh17: good idea !

shakeelmohamed:
Thanks for sticking with it @hjh17!

hjh17:
thanks for incredible guidance @shakeelmohamed !

PR #210, opened 2016-06-03
Shows error alert on mobile device 

### LINKED ISSUE #5: Display unsupported message for non-Android mobile users
TODO:

- [x] Try to get the best possible functionality on Android, autoplay will not work, by design.
- [ ] On iOS, and other mobile user agents display a message saying we don't support mobile browsers - since iOS will show videos in full screen, by design. see comments in #38 

Major bugs on Android

screenshot_2015-05-11-07-57-32
screenshot_2015-05-11-07-57-44

FleetAdmiral:
Shows an error alert if user is accessing the website on an mobile device , fixes #5 .

### DIFF
--- a/index.html
+++ b/index.html
@@ -39,7 +39,8 @@
 
         <hgroup>
             <h1>Zen Audio Player</h1>
-            <h2 class="color-grey" id="mobile-message">Listen to YouTube videos, without the distracting visuals.</h2>
+            <h2 class="color-grey">Listen to YouTube videos, without the distracting visuals.</h2>
+            <h2 class="color-grey" id="mobile-message"></h2>
         </hgroup>
     </header>
     <div id="container">
--- a/index.html
+++ b/index.html
@@ -39,11 +39,12 @@
 
         <hgroup>
             <h1>Zen Audio Player</h1>
-            <h2 class="color-grey" id="change-mobile">Listen to YouTube videos, without the distracting visuals.</h2>
+            <h2 class="color-grey" id="mobile-message">Listen to YouTube videos, without the distracting visuals.</h2>
         </hgroup>
     </header>
     <div id="container">
         <!-- The form -->
+
         <form id="form">
             <div class="input-group">
                 <input type="text" id="v" name="v" class="input-search color-grey" placeholder="Search, YouTube video ID, or URL">
--- a/index.html
+++ b/index.html
@@ -39,7 +39,7 @@
 
         <hgroup>
             <h1>Zen Audio Player</h1>
-            <h2 class="color-grey">Listen to YouTube videos, without the distracting visuals.</h2>
+            <h2 class="color-grey" id="change-mobile">Listen to YouTube videos, without the distracting visuals.</h2>
         </hgroup>
     </header>
     <div id="container">
--- a/js/everything.js
+++ b/js/everything.js
@@ -2,6 +2,7 @@
 
 // Pointer to Keen client
 var client;
+
 function anonymizeFileUrl() {
     var url = window.location.href;
     if (url.indexOf("file://") === 0) {
@@ -11,9 +12,6 @@ function anonymizeFileUrl() {
 }
 
 function sendKeenEvent(_msg, _data) {
-    if (!client) {
-        return;
-    }
     var d = {
         page_url: anonymizeFileUrl(), // eslint-disable-line camelcase
         user_agent: "${keen.user_agent}", // eslint-disable-line camelcase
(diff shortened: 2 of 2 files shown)


shakeelmohamed:
Thanks, can we actually do this for all mobile users? I'm not sure a website is even a reasonable means of using ZAP on mobile.

FleetAdmiral:
@shakeelmohamed Are you referring to android and the rest too? I thought the issue persisted only with iOS specific APIs. But i'm more than ok with taking up this task too. So what say , not support any mobile devices at all?

shakeelmohamed:
@FleetAdmiral correct, don't support any devices. Also can we make this a UI change? Around line 429 in `js/everything.js`, inside the onready listener, we can do the user agent check, use jQuery to do something like `$("#container").html("<p>We don't support mobile devices</p>");` then `return;`. The idea is to avoid even attempting to do any of the normal functionality when on a mobile device

FleetAdmiral:
@shakeelmohamed Sorry for being so late , kind of caught up with my own internship . Reading what you've written , i feel it wold be better if we just redirected the user to an error page mentioning/explaining that we don't support mobile , rather than changing the content of the same page. Gives us more space.
What do you say?

js/everything.js
```diff
@@ -474,7 +473,13 @@ $(function() {
             );
         }
     }
-
+    if ( /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent) )
```
shakeelmohamed: This needs to be moved up so it's the first thing done in the onready listener, before  the `// Keen.io` line.

js/everything.js
```diff
@@ -474,7 +473,13 @@ $(function() {
...
+    {
+        $("#container").ready(function() {
```
shakeelmohamed: Is this listener necessary? The document should already be ready

js/everything.js
```diff
@@ -474,7 +473,13 @@ $(function() {
...
+          $("#container").html("<p></p>");
```
shakeelmohamed: We shouldn't add a paragraph inside of a header. You can add a `<p>` tag to the HTML that's hidden by default, then here do `$("change-mobile").show();`

js/everything.js
FleetAdmiral: @shakeelmohamed Added the listener as a formality , will remove it and make the changes mentioned.

js/everything.js
```diff
@@ -474,7 +473,13 @@ $(function() {
...
+          $("#change-mobile").html("Whoops! Sorry , we don't support mobile devices.");
```
shakeelmohamed: I don't like this id name, something like `#mobile-message` is better

js/everything.js
```diff
@@ -474,7 +473,13 @@ $(function() {
...
+      });
+    }
```
shakeelmohamed: The last statement inside this `if` should be a `return;` we don't want to try doing anything else if we're on mobile.
We should also make sure the demo button is hidden at this point since it won't work

js/everything.js
FleetAdmiral: Will be done sorry for the trouble 👍

shakeelmohamed:
Left some more comments, please fix the style errors found on [Travis CI](https://travis-ci.org/zen-audio-player/zen-audio-player.github.io/jobs/136750025) as well

FleetAdmiral:
@shakeelmohamed All changes made 👍

js/everything.js
```diff
@@ -12,9 +12,6 @@ function anonymizeFileUrl() {
 }
 
 function sendKeenEvent(_msg, _data) {
-    if (!client) {
```
shakeelmohamed: revert this change

js/everything.js
```diff
@@ -430,13 +427,17 @@ function pickDemo() {
 }
 
 $(function() {
-    // Keen.io
-    if (typeof Keen !== "undefined") { // eslint-disable-line no-undef
```
shakeelmohamed: revert this change

js/everything.js
```diff
@@ -430,13 +427,17 @@ function pickDemo() {
...
-        client = new Keen({ // eslint-disable-line no-undef
-            projectId: "5690c384c1e0ab0c8a6c59c4",
-            writeKey: "630fa16847ce5ffb01c9cc00327498e4e7716e0f324fb14fdf0e83ffc06f9eacff5fad1313c2701efe4a91c88c34b8d8153cbb121c454056bb63caf60a46336dd9c9e9855ecc5202ef3151d798eda40896d5111f44005c707cbfb32c7ae31070d129d6f520d5604fdbce5ad31e9c7232"
-        });
+    if (/Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.usersAgent))
```
shakeelmohamed: After you revert the above change, move this `if ` statement above

js/everything.js
```diff
@@ -430,13 +427,17 @@ function pickDemo() {
...
+    {
```
shakeelmohamed: move this brace to end of the previous line

index.html
```diff
@@ -39,11 +39,12 @@
 
         <hgroup>
             <h1>Zen Audio Player</h1>
-            <h2 class="color-grey">Listen to YouTube videos, without the distracting visuals.</h2>
+            <h2 class="color-grey" id="mobile-message">Listen to YouTube videos, without the distracting visuals.</h2>
         </hgroup>
```
shakeelmohamed: There should be another empty `h2` here with `id="mobile-message"`

shakeelmohamed: Then add a CSS rule in `css/styles.css`:

```css
#mobile-message {
    display: none;
}
```

js/everything.js
```diff
@@ -430,13 +427,17 @@ function pickDemo() {
...
+        $("#container").hide();
+        $("#mobile-message").text("Whoops! Sorry , we don't support mobile devices.");
```
shakeelmohamed: Change the string to `Sorry, we don't support mobile devices.`

js/everything.js
```diff
@@ -430,13 +427,17 @@ function pickDemo() {
...
+        return;
     }
+    // Keen.io
```
shakeelmohamed: Remove  this

shakeelmohamed:
Looks like a merge went bad, left some comments

js/everything.js
FleetAdmiral: @shakeelmohamed Sorry , the new if statement and its body goes above the previous if statement ("if (typeof Keen !== "undefined") ") right?

shakeelmohamed: yes

PR #186, opened 2016-04-14
Add support for Soundcloud

monicacheung:
Closes #2

You can enter a Soundcloud URL/trackID in the search bar or as a query string (?v=) in the URL and ZAP will play it.

**Try these out in the ZAP search bar or ?v= in the URL:**
- https://soundcloud.com/theglitchmob/starve-the-ego-feed-the-soul
- https://api.soundcloud.com/tracks/3096795
- 3096795

There are changes to plyr.js in the directory bower_components/plyr to add support for Soundcloud to plyr. These are for test/debug until we send a PR to plyr.

**Todo/Things that broke:**
- [ ] Fix the tweet message.
- [ ] Fix storing current time of player and resume on returning to page.
- [ ] Update the UI a bit to be indicative of soundcloud support.
- [ ] More TODO in the code.

PR #212, opened 2016-06-09
Adds "Save Song" button to save music to localstorage.Still in need…

scoyne122:
Added a "save song" button that saves a song to local storage. Still needs a "my music" section that can display the users saved songs.

### DIFF
--- a/index.html
+++ b/index.html
@@ -82,7 +82,9 @@
             </div>
             <br>
             <div id="zen-video-description"></div>
+			<div id="save" class="btn">Save Song</div>
             <br>
+			<br>
         </div>
         <button id="demo" class="btn">
             <i class="fa fa-mouse-pointer"></i>
--- a/js/everything.js
+++ b/js/everything.js
@@ -272,7 +272,8 @@ var ZenPlayer = {
     			localStorage.setItem("mySongs", JSON.stringify([getCurrentVideoID()]));
     		}
     		else {
-    			localStorage.setItem("mySongs", JSON.stringify(mySongs.push(getCurrentVideoID())));
+    			mySongs.push(getCurrentVideoID());
+    			localStorage.setItem("mySongs", JSON.stringify(mySongs));
     		}
     	});
     }
--- a/js/everything.js
+++ b/js/everything.js
@@ -169,13 +169,14 @@ var ZenPlayer = {
                 that.setupTitle();
                 that.setupVideoDescription();
                 that.setupPlyrToggle();
+				that.setupSongSave();
             });
 
             plyrPlayer.addEventListener("playing", function() {
                 if (that.updated) {
                     return;
                 }
-
+				
                 // Start video from where we left off, if it makes sense
                 if (window.sessionStorage && window.sessionStorage.hasOwnProperty(videoID)) {
                     var resumeTime = window.sessionStorage[videoID];
@@ -263,6 +264,17 @@ var ZenPlayer = {
             toggleElement(event, ".plyr__video-wrapper", "Player");
         });
     },
+	setupSongSave: function() {
+		$("#save").click(function() {
+			var mySongs = JSON.parse(localStorage.getItem("mySongs"));
+			if (mySongs == null) {
+				localStorage.setItem("mySongs", JSON.stringify([getCurrentVideoID()]));
+			}
+			else {
+				localStorage.setItem("mySongs", JSON.stringify(mySongs.push(getCurrentVideoID())));
+			}
+		})
+	},
     getVideoDescription: function(videoID) {
         var description = "";
 
--- a/js/everything.js
+++ b/js/everything.js
@@ -176,7 +176,7 @@ var ZenPlayer = {
                 if (that.updated) {
                     return;
                 }
-				
+
                 // Start video from where we left off, if it makes sense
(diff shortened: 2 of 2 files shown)


js/everything.js
```diff
@@ -261,8 +262,21 @@ var ZenPlayer = {
...
+    		var mySongs = JSON.parse(localStorage.getItem("mySongs"));
+    		if (mySongs === null) {
+    			localStorage.setItem("mySongs", JSON.stringify([getCurrentVideoID()]));
+    		}
+    		else {
+    			mySongs.push(getCurrentVideoID());
+    			localStorage.setItem("mySongs", JSON.stringify(mySongs));
+    		}
+    	});
+    }
```
eirism: Insert a comma here to see if it builds

eirism: Insert a comma here and see if it builds

index.html
```diff
@@ -82,7 +82,9 @@ <h2 class="color-grey">Listen to YouTube videos, without the distracting visuals
             </div>
             <br>
             <div id="zen-video-description"></div>
+	    <div id="save" class="btn">Save Song</div>
             <br>
+	    <br>
```
shakeelmohamed: why this added break?

js/everything.js
```diff
@@ -169,6 +169,7 @@ var ZenPlayer = {
                 that.setupTitle();
                 that.setupVideoDescription();
                 that.setupPlyrToggle();
+		that.setupSongSave();
```
shakeelmohamed: tab indents

js/everything.js
```diff
@@ -261,8 +262,21 @@ var ZenPlayer = {
         // Show player button click event
         $("#togglePlayer").click(function(event) {
             toggleElement(event, ".plyr__video-wrapper", "Player");
+            	
         });
     },
+    setupSongSave: function() {
+    	$("#save").click(function() {
+    		var mySongs = JSON.parse(localStorage.getItem("mySongs"));
```
shakeelmohamed: tab

js/everything.js
```diff
@@ -261,8 +262,21 @@ var ZenPlayer = {
         // Show player button click event
         $("#togglePlayer").click(function(event) {
             toggleElement(event, ".plyr__video-wrapper", "Player");
+
```
shakeelmohamed: revert this change

js/everything.js
```diff
@@ -261,8 +262,21 @@ var ZenPlayer = {
         // Show player button click event
         $("#togglePlayer").click(function(event) {
             toggleElement(event, ".plyr__video-wrapper", "Player");
+            	
         });
     },
+    setupSongSave: function() {
+    	$("#save").click(function() {
+    		var mySongs = JSON.parse(localStorage.getItem("mySongs"));
+    		if (mySongs === null) {
```
shakeelmohamed: lots of tabs here 😄

shakeelmohamed:
@scoyne122 thanks for your efforts here. I think this raises an interesting idea, I've been thinking about doing something similar. The "right" way to implement this would be using the YouTube API (otherwise you'll lose your list when changing browsers or devices) to save videos to a playlist, and require oauth (see what I did for [YouTube watch later](https://github.com/shakeelmohamed/youtube-watch-later)). This also ties into #1 - adding support for YouTube playlists.

However, it's not quite that simple. When #186 goes in, we'll have Soundcloud support, meaning we'd need to support and cache 2 auth tokens. That's not necessarily a bad thing, but something we should collectively think about. Then we should consider how we want to handle playlists: a) YouTube & Soundcloud playlists are separated, or b) we add our own playlists feature that can contain music from either source. I think option b is more use friendly, but will be trickier to implement.

Let me know what you guys think! cc: @zen-audio-player/contributors

scoyne122:
@shakeelmohamed I agree with moving to a cloud based approach to ensure the user's music transfers over devices or browsers. However, assuming Sound Cloud is integrated, it may be easier to implement our own user login system rather than the youtube API with oauth. Perhaps we could create a PHP or node.js backend that connects to a mySQL?

Let me know your thoughts.

shakeelmohamed:
> Perhaps we could create a PHP or node.js backend that connects to a mySQL?

I've been against this so far in order to keep the project both free and totally open. What you see on the `master` branch of this repo is exactly what's served on https://ZenPlayer.Audio since we're using GitHub pages. Even if we implemented our own accounts, they would be based on oauth from another platform.  I'm not interested in adding a database for this project because of the added cost and complexity. If anyone needs to touch the backend or database there's additional security issues that will arise.

scoyne122:
Fair enough.
If we do decide on adding our own playlist feature (which I'm still a fan of), what would be your preferred method of doing so?

shakeelmohamed:
We could do something like having a reserved playlist name ("Zen Audio Player") on both YouTube and Soundcloud, then aggregate them on the website depending on which service(s) the user is authenticated.

YuriBrunetto:
Hey guys! 
@scoyne122 great work 👍 

But, as @shakeelmohamed said: I also think that it would be better to create a playlist on YouTube (maybe something with the name like `Yuri's ZenAudio Playlist`).

shakeelmohamed:
@scoyne122 shall we close this PR?

shakeelmohamed:
Closing since we've decided this won't be merged. Let's continue this conversation in #229

PR #226, opened 2016-07-21
Closes #110 Implementation of URI Library - Again

### LINKED ISSUE #110: Simplify all URL parsing code with URI.js
This seems like a good library and not too large of a dependency, what do we think? https://github.com/medialize/URI.js

rjstires:
Per Issue #100 request to implement URI.js throughout JS codebase for uniform parsing and manipulation of URIs. I chose to open this with URI.js as the other proposed library did not have mutation ability.

### Important

### Motivation and Context
- [X] If it fixes an open issue, include the text `Closes #1` (where 1 would be the issue number) to your commit message.

### Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)

### Description
- `parseYoutubeVideoID()` was updated to make use of URI.js both parsing and manipulation. I would like an example of slashes in the URI so that I might provide a better solution. (refer to TODO on `zap-common.js#74`)
- `isFileProtocol()` and `anonymizeFileUrl()` have been simplified by making use of the URI.js library.
- `getCurrentVideoID()` also makes use of the library. There was no the function wasn't self documenting and was vague in it's purpose. It seemed that it was ultimately attempting to locate the v query parameter, and if there were multiple, return the last value. This is how the function operates now.

### Final checklist:
- [x] I have updated the documentation accordingly.
- [ ] All tests passed.

I hate PRs, I'm hoping I did this right...

**Looking forward to feedback**


bower.json
```diff
@@ -20,6 +20,7 @@
     "keen-js": "~3.4.0",
     "trackjs": "~2.1.16",
     "typeahead.js": "~0.11.1",
-    "plyr": "1.6.x"
+    "plyr": "1.6.x",
+    "urijs": "uri.js#^1.18.1"
```
shakeelmohamed: This version number looks incorrect

js/everything.js
```diff
@@ -364,56 +371,71 @@ function toggleElement(event, toggleID, buttonText) {
...
-        v = vParams[vParams.length - 1].replace("v=", "");
+    var v = URI(window.location).search(true).v;
+    if (!v) {
+        return null;
     }
-    else if (v.length > 1) {
-        return wrapParseYouTubeVideoID(v);
+
+    if (Array.isArray(v)) {
+        v = v[0];
```
shakeelmohamed: We wan't the last one, right?

js/everything.js
```diff
@@ -364,56 +371,71 @@ function toggleElement(event, toggleID, buttonText) {
     }
 }
 
+/**
+ * wrapParseYouTubeVideoID the first v query value.
+ * Will return null if there's no v query param
+ * @return {string|null}
+ */
 function getCurrentVideoID() {
```
shakeelmohamed: If possible, I'd like to see this written with a single `return`. If it feels non-trivial we can do it later

js/everything.js
```diff
@@ -364,56 +371,71 @@ function toggleElement(event, toggleID, buttonText) {
...
+        v = v[0];
     }
-    return v;
+
+    return wrapParseYouTubeVideoID(v);
 }
 
+/**
+ * Return the current times position, provided by the t query param, parsed to seconds.
+ * @returns {*}
```
shakeelmohamed: Should return a `Number`

js/everything.js
```diff
@@ -364,56 +371,71 @@ function toggleElement(event, toggleID, buttonText) {
...
+    var t = parseInt(URI(window.location).search(true).t, 10);
     if (t > 0 && t < Number.MAX_VALUE) {
         return t;
     }
     return 0;
 }
-
+/**
+ * Return the q query param if one exists.
+ * @returns {boolean|string|*|Array}
```
shakeelmohamed: Return value is String, possibly a bool/null depending on what we get back from the URI library

js/everything.js
```diff
@@ -364,56 +371,71 @@ function toggleElement(event, toggleID, buttonText) {
...
-    var url = removeSearchQueryFromURL(window.location.href);
-    // Remove any #s which break functionality
-    url = url.replace("#", "");
-    url += "?v=" + videoID;
+    var url = URI(window.location);
+    url.setSearch("v", videoID);
+
     if (videoPosition) {
-        url += "&t=" + videoPosition;
+        url.addSearch("t", videoPosition);
```
shakeelmohamed: why not `setSearch("t", videoPosition)`? is there any different between those 2 functions?

js/everything.js
```diff
@@ -364,56 +371,71 @@ function toggleElement(event, toggleID, buttonText) {
...
     }
-    return url;
+
+    return url.toString().replace("#", "");
```
shakeelmohamed: Does the URI library have something like `sethash()` we can use instead?

js/everything.js
```diff
@@ -364,56 +371,71 @@ function toggleElement(event, toggleID, buttonText) {
...
+ */
 function makeSearchURL(searchQuery) {
-    var url = removeSearchQueryFromURL(window.location.href);
-    // Remove any #s which break functionality
-    url = url.replace("#", "");
-
-    return url + "?q=" + encodeURIComponent(searchQuery);
+    var url = URI(window.location);
+    url.setSearch("q", searchQuery);
+    return url.toString().replace("#", "");
```
shakeelmohamed: same as above, can we use a URI function for this?

js/everything.js
rjstires: I've decided to refactor the 'cleaning' of strings into a new function in the next commit.

js/everything.js
rjstires: See new `cleanURL` function.

js/everything.js
rjstires: See new `cleanURL` function.

js/everything.js
rjstires: Documentation updated.

js/everything.js
rjstires: Documentation updated.

js/everything.js
rjstires: Fixed.

rjstires:
Just a note on the dependency version for urijs, I'm largely unfamiliar with bower, but the version was put in as it was in the original commit. I changed it to reflect, what appears to be, standard convention. YMMV

js/everything.js
```diff
@@ -378,20 +378,21 @@ function toggleElement(event, toggleID, buttonText) {
...
 function getCurrentVideoID() {
     var v = URI(window.location).search(true).v;
-    if (!v) {
-        return null;
-    }
-
-    if (Array.isArray(v)) {
-        v = v[0];
+    var r;
+    if (v) {
```
shakeelmohamed: Is this outer if necessary?

js/everything.js
```diff
@@ -378,20 +378,21 @@ function toggleElement(event, toggleID, buttonText) {
  */
 function getCurrentVideoID() {
     var v = URI(window.location).search(true).v;
-    if (!v) {
-        return null;
-    }
-
-    if (Array.isArray(v)) {
-        v = v[0];
+    var r;
```
shakeelmohamed: Since only 1 possible if/else can be executed, we can just update the `v` variable - we don't need `r` at all.

js/everything.js
```diff
@@ -378,20 +378,21 @@ function toggleElement(event, toggleID, buttonText) {
...
     }
 
-    return wrapParseYouTubeVideoID(v);
+    return r;
 }
 
 /**
  * Return the current times position, provided by the t query param, parsed to seconds.
- * @returns {*}
+ * @returns {number}
```
shakeelmohamed: nit: `Number` since we're stating the type

js/everything.js
```diff
@@ -433,9 +436,20 @@ function makeListenURL(videoID, videoPosition) {
...
+      .setSearch("q", searchQuery)
+      .toString();
+}
+
+/**
+ * Remove any search params or fragments from a URL.
+ * @param url
+ * @returns {string} The stripped URL
+ */
+function cleanURL(url) {
```
shakeelmohamed: Maybe we just want to call `fragment("")`, do we need to clear the query args (assuming that's what `search("")`)?

Also, this should be moved above the first usage. Some browsers may error if it's used before it's declared. I know Firefox did this the last time I checked

js/zap-common.js
```diff
@@ -76,5 +81,4 @@ function parseYoutubeVideoID(url) {
 
         return videoInfo;
     }
-}
-/* eslint-enable */
\ No newline at end of file
+}/* eslint-enable */
```
shakeelmohamed: revert this removed newline

js/everything.js
rjstires: This appears to be an issue with Github, the code in file is;
`function getCurrentVideoID() {
    var v = URI(window.location).search(true).v;
    var r;
    if (v) {
        if (Array.isArray(v)) {
            r = wrapParseYouTubeVideoID(v.pop());
        } else if (v) {
            r = wrapParseYouTubeVideoID(v);
        }
    }

    return r;
}`

js/zap-common.js
```diff
@@ -39,32 +34,42 @@ function getYouTubeVideoDescription(videoID, youTubeDataApiKey, onSuccess, onFai
...
+            videoInfo.id = url.path().slice(1);
         }
         // Assume YouTube video ID string
         else {
             videoInfo.format = "video ID";
-            videoInfo.id = url;
+            videoInfo.id = uri;
         }
 
         var slashPos = videoInfo.id.indexOf("/");
```
shakeelmohamed: Does this case get handled by the URI library? If so, we can remove these 4 lines

js/everything.js
```diff
@@ -364,56 +371,85 @@ function toggleElement(event, toggleID, buttonText) {
...
-    var v = getParameterByName(window.location.search, "v");
-    // If the URL had 2 v parameters, try parsing the second (usually when ?v=someurl&v=xyz)
-    var vParams = window.location.search.match(/v=\w+/g);
-    if (vParams && vParams.length > 1) {
-        v = vParams[vParams.length - 1].replace("v=", "");
-    }
-    else if (v.length > 1) {
-        return wrapParseYouTubeVideoID(v);
+    var v = URI(window.location).search(true).v;
+    var r;
```
shakeelmohamed: @rjstires Here, re: https://github.com/zen-audio-player/zen-audio-player.github.io/pull/226#discussion_r71813992

shakeelmohamed:
Closing in favor of #245!

PR #232, opened 2016-07-30
Thumbnails In Search Results (No Thumbnail When Playing)

artforlife:
The search results are now displayed with thumbnails. No image is displayed when a video is selected for playing.
### Important

### Motivation and Context
- [ ] If it fixes an open issue, include the text `Closes #1` (where 1 would be the issue number) to your commit message.
### Types of changes

### Description
### Final checklist:

- [x] All tests passed.

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -141,14 +141,13 @@ a {
     margin: 0 auto;
     clear: left;
     float: left;
-    margin: 5px;
 }
 
 #search-results h4 a img {
-	vertical-align: middle;
-	border-radius: 8px;
-	margin-right: 5px;
-	padding: 2px;
+    vertical-align: middle;
+    border-radius: 8px;
+    margin-right: 5px;
+    padding: 2px;
 }
 
 
--- a/css/styles.css
+++ b/css/styles.css
@@ -137,7 +137,20 @@ a {
 
 #search-results li {
     list-style-type: none;
+    display: block;
+    margin: 0 auto;
+    clear: left;
+    float: left;
+    margin: 5px;
 }
+
+#search-results h4 a img {
+	vertical-align: middle;
+	border-radius: 8px;
+	margin-right: 5px;
+	padding: 2px;
+}
+
 
 /* Input field fixes and search suggestion style.
  * typeahead conflicts with primercss */
--- a/css/styles.css
+++ b/css/styles.css
@@ -143,7 +143,7 @@ a {
     float: left;
 }
 
-#search-results h4 a img {
+#search-results a img {
     vertical-align: middle;
     border-radius: 8px;
     margin-right: 5px;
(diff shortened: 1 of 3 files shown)


css/styles.css
```diff
@@ -137,8 +137,20 @@ a {
...
 }
 
+#search-results a img {
+    vertical-align: middle;
+    border-radius: 8px;
+    margin-right: 5px;
+    padding: 2px;
+}
+
+
```
shakeelmohamed: nit: extra blank line

js/everything.js
```diff
@@ -495,11 +495,11 @@ $(function() {
                     $("#search-results").show();
                     // Clear out results
                     $("#search-results ul").html("");
-
-                    var start = "<li><h4><a href=?v=";
+                    var start = "<li><h4><a style='float:right' href=?v=";
```
shakeelmohamed: I don't like using single quotes in HTML, let's update all of these

js/everything.js
```diff
@@ -495,11 +495,11 @@ $(function() {
...
                     var end = "</a></h4></li>";
                     $.each(data.items, function(index, result) {
-                        $("#search-results ul").append(start + result.id.videoId + ">" + result.snippet.title  + end);
+                        var thumbTag = "<img style='float:left; margin: auto' src='https://img.youtube.com/vi/" + result.id.videoId + "/1.jpg'>";
```
shakeelmohamed: I'd like to break this up into a couple of steps, make the image URL it's own var, etc to make this more readable.

js/everything.js
shakeelmohamed: Let's remove all the inline style and create a new CSS class, same for the results themselves

shakeelmohamed:
Can you take a screenshot with before/after?

artforlife:
Here is a screenshot. I shall clean up the code soon.
screen shot 2016-07-30 at 9 40 35 am

shakeelmohamed:
Thanks, it definitely disturbs the current layout. We might want to use smaller thumbnail images (there may be a different URL for different sizes of the same thumbnail), and probably left-align the video title so it looks more like YouTube - see the screenshot below:

capturfiles_2

artforlife:
I can left-align the title without a problem. How much smaller do we want the thumbnails? They are already quite tiny. Also, I do not see how they disturb the layout unless I simply misunderstand what you mean.

shakeelmohamed:
> How much smaller do we want the thumbnails?

Let's try 50% of how they are now, can we make sure the aspect ratio is correct as well? For ABGT200 (second result in both screenshots), in your implementation there are horizontal black bars unlike the YouTube search results.

> Also, I do not see how they disturb the layout unless I simply misunderstand what you mean.

The video titles are center aligned, but also pushed to the right by X pixels (where X is the width of the thumbnail).

artforlife:
I was away for awhile. Will do a PR in next couple days. Also back on the Electron side!

shakeelmohamed:
@artforlife thanks for the update!

shakeelmohamed:
Closing due to inactivity

PR #234, opened 2016-07-31
Anchor timestamp URLs in description. Closes #228

### LINKED ISSUE #228: Show Track Timemarks If Original Video Has Them in Description
For instance, this video has track names and their start times in its description. It would be neat to show these once the user starts playback of the video.

https://www.youtube.com/watch?v=PEQEOP1zh_c

hogniRunar:
Provide a general summary of your changes in the **Title** above.

### Important

### Motivation and Context
- [x] If it fixes an open issue, include the text `Closes #1` (where 1 would be the issue number) to your commit message.

### Types of changes

### Description

### Final checklist:
- [x] All tests passed.

This closes the issue #228. The main changes that were made were:
Pass videoID into the setupVideoDescription funciton (to be able to construct the URLs)
Two new functions were created called anchorTimestamps and convertTimestamps. The first creates URLs in the description and the latter converts timestamps from 01:00:00 to 01h00m00s for example

### DIFF
--- a/js/everything.js
+++ b/js/everything.js
@@ -175,7 +175,7 @@ var ZenPlayer = {
 
                 // Initialize UI
                 that.setupTitle();
-                that.setupVideoDescription();
+                that.setupVideoDescription(videoID);
                 that.setupPlyrToggle();
             });
 
@@ -265,8 +265,9 @@ var ZenPlayer = {
         $("#zen-video-title").html(tmpVideoTitle);
         $("#zen-video-title").attr("href", this.videoUrl);
     },
-    setupVideoDescription: function() {
+    setupVideoDescription: function(videoID) {
         var description = anchorURLs(this.videoDescription);
+        description = anchorTimestamps(description, videoID);
         $("#zen-video-description").html(description);
         $("#zen-video-description").hide();
 
@@ -425,6 +426,30 @@ function anchorURLs(text) {
     var re = /((?:http|https)\:\/\/[a-zA-Z0-9\-\.]+\.[a-zA-Z]{2,3}(?:\/\S*[^\.\s])?)/g;
     /* Wraps all found URLs in <a> tags */
     return text.replace(re, "<a href=\"$1\" target=\"_blank\">$1</a>");
+}
+
+function anchorTimestamps(text, videoID) {
+    /* RegEx to match
+      hh:mm:ss
+      h:mm:ss
+      mm:ss
+      m:ss
+    and wraps the timestamps in <a> tags
+    */
+    var re = /((?:[0-5]\d|\d|)(?:\d|\:[0-5]\d)(?:$|\:[0-5]\d))/g;
+    return text.replace(re, (match, $1) => "<a href=\"" + makeListenURL(videoID, convertTimestamp($1)) + "\">" + $1 + "</a>");
+}
+
+function convertTimestamp(stamp) {
+    // changes timestamps from 01:00:00 to 01h00m00s for example for the makeListenURL function
+    if (stamp.substr(0, stamp.indexOf(":")) !== "00") {
+        stamp = stamp.replace(":", "h");
+    } else {
+        stamp = stamp.slice(3, stamp.length);
+    }
+    stamp = stamp.replace(":", "m");
+    stamp += "s";
+    return stamp;
 }
 
 function wrapParseYouTubeVideoID(url) {
--- a/js/everything.js
+++ b/js/everything.js
@@ -456,7 +456,7 @@ function convertTimestamp(timestamp) {
         minutes += parseInt(timeComponents[1], 10) * 60; // convert minutes to seconds
         seconds += parseInt(timeComponents[2], 10); // add remaining seconds
     } else {
-        minutes += parseInt(timeComponents[0], 10) * 60; // convert minutes to sec§onds
+        minutes += parseInt(timeComponents[0], 10) * 60; // convert minutes to seconds
         seconds += parseInt(timeComponents[1], 10); // add remaining seconds
     }
     return hours + minutes + seconds;
--- a/js/everything.js
+++ b/js/everything.js
@@ -453,12 +453,12 @@ function convertTimestamp(timestamp) {
     var numberOfColons = timestamp.match(/:/g).length;
     timestamp = timestamp.split(":");
     if (numberOfColons === 2) {
-        hours += parseInt(timestamp[0]) * 60 * 60; // convert hours to seocnds
-        minutes += parseInt(timestamp[1]) * 60; // convert minutes to seconds
-        seconds += parseInt(timestamp[2]); // add remaining seconds
+        hours += parseInt(timestamp[0], 10) * 60 * 60; // convert hours to seocnds
+        minutes += parseInt(timestamp[1], 10) * 60; // convert minutes to seconds
+        seconds += parseInt(timestamp[2], 10); // add remaining seconds
     } else {
-        minutes += parseInt(timestamp[0]) * 60; // convert minutes to sec§onds
-        seconds += parseInt(timestamp[1]); // add remaining seconds
+        minutes += parseInt(timestamp[0], 10) * 60; // convert minutes to sec§onds
+        seconds += parseInt(timestamp[1], 10); // add remaining seconds
     }
     return hours + minutes + seconds;
 }
--- a/js/everything.js
+++ b/js/everything.js
@@ -450,9 +450,8 @@ function convertTimestamp(timestamp) {
     var seconds = 0;
     var minutes = 0;
     var hours = 0;
-    var numberOfColons = timestamp.match(/:/g).length;
     var timeComponents = timestamp.split(":");
-    if (numberOfColons === 2) {
+    if (timeComponents.length === 3) {
         hours += parseInt(timeComponents[0], 10) * 60 * 60; // convert hours to seocnds
         minutes += parseInt(timeComponents[1], 10) * 60; // convert minutes to seconds
         seconds += parseInt(timeComponents[2], 10); // add remaining seconds
--- a/js/everything.js
+++ b/js/everything.js
@@ -441,20 +441,26 @@ function anchorTimestamps(text, videoID) {
     (?:$|\:[0-5]\d)) either the string ends or is a a number between 00-59
     */
     var re = /((?:[0-5]\d|\d|)(?:\d|\:[0-5]\d)(?:$|\:[0-5]\d))/g;
-    return text.replace(re, function(match) { return "<a href=\"" + makeListenURL(videoID, convertTimestamp(match)) + "\">" + match + "</a>"; });
+    return text.replace(re, function(match) {
+        return "<a href=\"" + makeListenURL(videoID, convertTimestamp(match)) + "\">" + match + "</a>";
+    });
 }
 
-function convertTimestamp(stamp) {
-    var numberOfColons = stamp.match(/:/g).length;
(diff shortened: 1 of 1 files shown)


js/everything.js
```diff
@@ -427,6 +428,30 @@ function anchorURLs(text) {
...
+function anchorTimestamps(text, videoID) {
+    /* RegEx to match
+      hh:mm:ss
+      h:mm:ss
+      mm:ss
+      m:ss
+    and wraps the timestamps in <a> tags
+    */
+    var re = /((?:[0-5]\d|\d|)(?:\d|\:[0-5]\d)(?:$|\:[0-5]\d))/g;
+    return text.replace(re, (match, $1) => "<a href=\"" + makeListenURL(videoID, convertTimestamp($1)) + "\">" + $1 + "</a>");
```
shakeelmohamed: I'd rather not use `=>` syntax, since I don't believe we're using it anywhere else

js/everything.js
```diff
@@ -427,6 +428,30 @@ function anchorURLs(text) {
...
 
+function anchorTimestamps(text, videoID) {
+    /* RegEx to match
+      hh:mm:ss
+      h:mm:ss
+      mm:ss
+      m:ss
+    and wraps the timestamps in <a> tags
+    */
+    var re = /((?:[0-5]\d|\d|)(?:\d|\:[0-5]\d)(?:$|\:[0-5]\d))/g;
```
shakeelmohamed: This regex is a bit painful to read without a long comment explaining each part of it. I might like to see this broken up into multiple regexes for readability

js/everything.js
```diff
@@ -427,6 +428,30 @@ function anchorURLs(text) {
...
+    return text.replace(re, (match, $1) => "<a href=\"" + makeListenURL(videoID, convertTimestamp($1)) + "\">" + $1 + "</a>");
+}
+
+function convertTimestamp(stamp) {
+    // changes timestamps from 01:00:00 to 01h00m00s for example for the makeListenURL function
```
shakeelmohamed: The logic here is wrong. The `t` querystring parameter is the time in seconds, ie: `t=120` for 2 minutes. The return value of this function should be an integer

hogniRunar:
changed the code a bit. the links work now and the return text.replace() has a function as the second parameter to construct the anchor string

js/everything.js
```diff
@@ -435,21 +435,26 @@ function anchorTimestamps(text, videoID) {
...
+    ((?:[0-5]\d|\d|) either the string is "colon 00-59" or "0-9" or "blank"
+    (?:\d|\:[0-5]\d) either the string is "colon 0-9" or "colon 00-59"
+    (?:$|\:[0-5]\d)) either the string ends or is a a number between 00-59
     */
     var re = /((?:[0-5]\d|\d|)(?:\d|\:[0-5]\d)(?:$|\:[0-5]\d))/g;
-    return text.replace(re, (match, $1) => "<a href=\"" + makeListenURL(videoID, convertTimestamp($1)) + "\">" + $1 + "</a>");
+    return text.replace(re, function(match) { return "<a href=\"" + makeListenURL(videoID, convertTimestamp(match)) + "\">" + match + "</a>"; });
 }
 
 function convertTimestamp(stamp) {
```
shakeelmohamed: can we use a different variable name here? I don't like `stamp`

js/everything.js
```diff
@@ -435,21 +435,26 @@ function anchorTimestamps(text, videoID) {
...
       m:ss
     and wraps the timestamps in <a> tags
+    RegEx explanation:
+    ((?:[0-5]\d|\d|) either the string is "colon 00-59" or "0-9" or "blank"
+    (?:\d|\:[0-5]\d) either the string is "colon 0-9" or "colon 00-59"
+    (?:$|\:[0-5]\d)) either the string ends or is a a number between 00-59
     */
     var re = /((?:[0-5]\d|\d|)(?:\d|\:[0-5]\d)(?:$|\:[0-5]\d))/g;
-    return text.replace(re, (match, $1) => "<a href=\"" + makeListenURL(videoID, convertTimestamp($1)) + "\">" + $1 + "</a>");
+    return text.replace(re, function(match) { return "<a href=\"" + makeListenURL(videoID, convertTimestamp(match)) + "\">" + match + "</a>"; });
```
shakeelmohamed: Can we do something like this? I'd rather not use a function in `.replace()` since it's a bit clunky

```javascript
text.replace(re, "<a href=\"" + makeListenURL(videoID, convertTimestamp("$1")) + "\">$1</a>");
```

js/everything.js
```diff
@@ -435,21 +435,26 @@ function anchorTimestamps(text, videoID) {
...
+    if (numberOfColons === 2) {
+        seconds = (+stamp[0]) * 60 * 60 + (+stamp[1]) * 60 + (+stamp[2]);
     } else {
-        stamp = stamp.slice(3, stamp.length);
+        seconds = (+stamp[0]) * 60 + (+stamp[1]);
     }
-    stamp = stamp.replace(":", "m");
-    stamp += "s";
-    return stamp;
+    console.log(seconds);
```
shakeelmohamed: let's remove this, eventually.

js/everything.js
```diff
@@ -435,21 +435,26 @@ function anchorTimestamps(text, videoID) {
...
 
 function convertTimestamp(stamp) {
-    // changes timestamps from 01:00:00 to 01h00m00s for example for the makeListenURL function
-    if (stamp.substr(0, stamp.indexOf(":")) !== "00") {
-        stamp = stamp.replace(":", "h");
+    var numberOfColons = stamp.match(/:/g).length;
+    var seconds = 0;
+    stamp = stamp.split(":");
+    if (numberOfColons === 2) {
+        seconds = (+stamp[0]) * 60 * 60 + (+stamp[1]) * 60 + (+stamp[2]);
```
shakeelmohamed: Is the `+` in `+stamp[0]` trying to convert a substring to an integer? We should user `parseInt(foo, 10)` instead

js/everything.js
```diff
@@ -435,21 +435,26 @@ function anchorTimestamps(text, videoID) {
...
 }
 
 function convertTimestamp(stamp) {
-    // changes timestamps from 01:00:00 to 01h00m00s for example for the makeListenURL function
-    if (stamp.substr(0, stamp.indexOf(":")) !== "00") {
-        stamp = stamp.replace(":", "h");
+    var numberOfColons = stamp.match(/:/g).length;
+    var seconds = 0;
+    stamp = stamp.split(":");
+    if (numberOfColons === 2) {
```
shakeelmohamed: For super clarity, I'd actually like to see this broken up into `hours`, `minutes`, `seconds` then add/multiply everything up at the end of the function

js/everything.js
```diff
@@ -427,6 +428,35 @@ function anchorURLs(text) {
...
+      mm:ss
+      m:ss
+    and wraps the timestamps in <a> tags
+    RegEx explanation:
+    ((?:[0-5]\d|\d|) either the string is "colon 00-59" or "0-9" or "blank"
+    (?:\d|\:[0-5]\d) either the string is "colon 0-9" or "colon 00-59"
+    (?:$|\:[0-5]\d)) either the string ends or is a a number between 00-59
+    */
+    var re = /((?:[0-5]\d|\d|)(?:\d|\:[0-5]\d)(?:$|\:[0-5]\d))/g;
+    return text.replace(re, function(match) { return "<a href=\"" + makeListenURL(videoID, convertTimestamp(match)) + "\">" + match + "</a>"; });
```
shakeelmohamed: Re-commenting to make sure we don't forget about this https://github.com/zen-audio-player/zen-audio-player.github.io/pull/234#discussion_r73083405

js/everything.js
hogniRunar: yeah that's what I tried before and that just passes the string "$1" into the convertTimestamp function. I am having trouble getting the actual match sent into the convertTimestamp function.

shakeelmohamed: Okay, I can tweak later once I get some time. Let's at least put the function body on it's own line (and indented by 4 spaces) so it's a bit more readable

js/everything.js
```diff
@@ -441,20 +441,26 @@ function anchorTimestamps(text, videoID) {
...
+function convertTimestamp(timestamp) {
     var seconds = 0;
-    stamp = stamp.split(":");
+    var minutes = 0;
+    var hours = 0;
+    var numberOfColons = timestamp.match(/:/g).length;
+    timestamp = timestamp.split(":");
     if (numberOfColons === 2) {
-        seconds = (+stamp[0]) * 60 * 60 + (+stamp[1]) * 60 + (+stamp[2]);
+        hours += parseInt(timestamp[0]) * 60 * 60; // convert hours to seocnds
```
shakeelmohamed: For every call to `parseInt()` we should pass in the second parameter  (radix) as `10` to avoid potential issues where the value can be interpreted as base-10

js/everything.js
```diff
@@ -427,6 +428,41 @@ function anchorURLs(text) {
...
+    return text.replace(re, function(match) {
+        return "<a href=\"" + makeListenURL(videoID, convertTimestamp(match)) + "\">" + match + "</a>";
+    });
+}
+
+function convertTimestamp(timestamp) {
+    var seconds = 0;
+    var minutes = 0;
+    var hours = 0;
+    var numberOfColons = timestamp.match(/:/g).length;
```
shakeelmohamed: Why do we need this? Can't we just use the value of `timestamp.split(":").length`?

js/everything.js
```diff
@@ -427,6 +428,41 @@ function anchorURLs(text) {
...
+    timestamp = timestamp.split(":");
```
shakeelmohamed: I'd make a new var here, since `timestamp` looks confusing. Maybe `timeComponents`

hogniRunar: yeah, agreed

js/everything.js
hogniRunar: Yeah that works. Looks a lot better

js/everything.js
```diff
@@ -427,6 +428,40 @@ function anchorURLs(text) {
...
+    });
+}
+
+function convertTimestamp(timestamp) {
+    var seconds = 0;
+    var minutes = 0;
+    var hours = 0;
+    var timeComponents = timestamp.split(":");
+    if (timeComponents.length === 3) {
+        hours += parseInt(timeComponents[0], 10) * 60 * 60; // convert hours to seocnds
```
shakeelmohamed: None of these need to be `+=` now

js/everything.js
```diff
@@ -427,6 +428,40 @@ function anchorURLs(text) {
...
+    var seconds = 0;
+    var minutes = 0;
+    var hours = 0;
+    var timeComponents = timestamp.split(":");
+    if (timeComponents.length === 3) {
+        hours = parseInt(timeComponents[0], 10) * 60 * 60; // convert hours to seocnds
+        minutes = parseInt(timeComponents[1], 10) * 60; // convert minutes to seconds
+        seconds = parseInt(timeComponents[2], 10); // add remaining seconds
+    } else {
+        minutes = parseInt(timeComponents[0], 10) * 60; // convert minutes to sec§onds
```
shakeelmohamed: I just noticed you have a funny character in the comment here`§`. Once that's fixed I'll do some manual testing before merging

hogniRunar:
Roger, sounds good. Thanks for all your comments

shakeelmohamed:
Works great, thanks so much @hogniRunar! Welcome to the org

hogniRunar:
Thanks! :)

PR #236, opened 2016-08-04
Fix always hidden "show/hide description" button.

shakeelmohamed:
@hogniRunar I have no idea how you tested the timestamped URL feature without this 💀


hogniRunar:
yeah it was a bit annoying. I "Inspect"-ed the "Show Player" button and then found the show description button in the html and toggled the display: none in the CSS. Pretty stupid way but I just wanted to focus on this specific issue :)

Issue #225, opened 2016-07-19
Video IDs containing - don't parse correctly


shakeelmohamed:
For video URLs like ` https://www.youtube.com/watch?v=UAWcs5H-qgQ` only `UAWcs5H` is parsed out. The bug is likely somewhere in `js/zap-common.js`.

Also:

* `https://www.youtube.com/watch?v=i-b4wn2TPRI` only `i` is parsed

cipher1729:
Both the above URLs seem to work just fine.  Could you please elaborate on how to reproduce this issue?

shakeelmohamed:
@cipher1729 sure, go to either of these URLs:
* https://zenplayer.audio/?v=https://www.youtube.com/watch?v=UAWcs5H-qgQ
* https://zenplayer.audio/?v=https://www.youtube.com/watch?v=i-b4wn2TPRI

cipher1729:
I've found a fix for it.  I'll create a pull request

rjstires:
This would inherently be addressed by PR #110

rjstires:
Will be closed by #226

PR #241, opened 2016-09-06
Fixed the Video ID Parse. Issue #225

pedrojanula:
Fixed the bug. Now, the video ID will parse correctly when inserting into the URL as in this example:

https://zenplayer.audio/?v=https://www.youtube.com/watch?v=UAWcs5H-qgQ


shakeelmohamed:
Thanks for the effort @pedrojanula, this actually didn't fix the bug... but it did show me where to look! See #243

PR #249, opened 2017-04-21
WIP: reverts to Home when form is empty 

xianrusso:
Hey!  It should work, but please let me know if you encounter any flaws.  

I am targeting typeahead.js default classes since I didn’t want to create any new classes and there weren’t any noticeable attributes or properties to attach to.  Using simple jQuery that you have in the document, the suggestions should hide when there isn’t text in the form.

However, when I tried to test locally, I was unable to retrieve any data through the API.  I am unsure as to whether it is a problem with my text editor, my local environment, or pizza fairies, but the main reason that I am stating this is because:
I was unable to test it.

So, that is why I’ve been using the word “should”.  

Anyways, just let me know if there is anything to be fixed!

### DIFF
--- a/js/everything.js
+++ b/js/everything.js
@@ -651,6 +651,11 @@ $(function() {
         $("#demo").hide();
     }
 
+    // Reverts to Home when there is no text in input
+    if ($(".tt-menu").hasClass("tt-empty")) {
+        $(".tt-menu > .tt-dataset").hide();
+    }
+
     // Handle demo link click
     $("#demo").click(function(event) {
         event.preventDefault();


js/everything.js
```diff
@@ -651,6 +651,11 @@ $(function() {
         $("#demo").hide();
     }
 
+    // Reverts to Home when there is no text in input
+    if ($(".tt-menu").hasClass("tt-empty")) {
```
shakeelmohamed: This logic is correct in theory, but we need to do the check inside an event handler when the text in `.tt-menu` changes. I'll let you figure this one out, it might end up looking something like:

```js
$(".tt-menu").on("input", function(e) {
    // Move your logic here
});
```

You should definitely be able to test locally, read `CONTRIBUTING.md` for setup steps.

xianrusso:
screen shot 2017-04-24 at 2 18 55 pm
screen shot 2017-04-24 at 2 19 09 pm

Here are the errors my terminal is throwing.  I have the latest version of node installed, and I also uninstalled and reinstalled everything as well.  The issue seems to lie in the linter, but that is a very surface-level diagnosis.  

After typing "npm bugs zen-audio-player", this is the error that _might_ be causing the entire problem: 
screen shot 2017-04-24 at 2 22 20 pm

The good news is that the code submitted has thrown zero errors (minus a trivial indentation error that I will fix)!  

Let me know what I can do to help solve these errors though!

shakeelmohamed:
@xianrusso If you fix the indentation error you shouldn't see any errors. The entire `npm test` command will fail if any of `test` or `pretest` commands fail. You can look in `package.json` to see what actually gets run for those steps

xianrusso:
@shakeelmohamed & you were exactly right!  Zero errors were thrown, every test was passed. Thank you for being patient!

shakeelmohamed:
@xianrusso Just took a look here, this doesn't work for me - please take a look

shakeelmohamed:
Closing due to inactivity

PR #254, opened 2017-07-21
Fix video resume

avalan4e57:
The fix solves the problem of resuming video from where we left it. 

### Motivation and Context
- [x] Previously it didn't work because videoDuration variable was initiated with 0. I guess that's because of differences in API loading speed (plyr started playing before all information from youtube arrived, in this case videoDuration option).

### Types of changes
- [x] Bug fix (non-breaking change which fixes an issue). Just added two lines of code that make the player not to start before videoDuration option arrives from youtube.

### Description
- [x] When videoDuration option is got with `var videoDuration = plyrPlayer.plyr.embed.getDuration();` next follows a simple piece of code `if (videoDuration === 0) { return; }`

### Final checklist:
- [x] All tests passed.


shakeelmohamed:
Closing as dupe of #255

PR #255, opened 2017-07-22
Start with youtube time

avalan4e57:
This fix is about to start the playing time from the place it was on YouTube video when we left.

### Motivation and Context
- [x] The change is required because YouTube today gives time option with `?t=` but the code works only with `&t=`. Now it works with both.
- [x] This fix is based on my previous commit which is "Fix video resume". That one fixes the problem of resuming video from where we left in on zenplayer.audio.

### Types of changes

### Description
- [x] Changed the line `var formValueTime = /&t=(\d*)$/g.exec(formValue);` to `var formValueTime = /[?&]t=(\d*)$/g.exec(formValue);` The changes are at the very beginning of the RegExp (`&t` is replaced with `[?&]t`)
- [x] Added two similar pieces of code `if (videoDuration === 0) { return; }` after videoDuration variable is initiated with `var videoDuration = plyrPlayer.plyr.embed.getDuration();` This change is required because previously the videoDuration variable was initiating with 0 most of the time. My guess about that problem is different loading speed of APIs. When plyr API starts playing the YouTube API still haven't got all required options (like videoDuration in the case).
  

### Final checklist:
- [x] All tests passed.

### DIFF
--- a/js/everything.js
+++ b/js/everything.js
@@ -185,17 +185,14 @@ var ZenPlayer = {
             });
 
             plyrPlayer.addEventListener("playing", function() {
-                if (that.updated) {
+                var videoDuration = plyrPlayer.plyr.embed.getDuration();
+                if (that.updated || videoDuration === 0) {
                     return;
                 }
 
                 // Start video from where we left off, if it makes sense
                 if (window.sessionStorage && window.sessionStorage.hasOwnProperty(videoID)) {
                     var resumeTime = window.sessionStorage[videoID];
-                    var videoDuration = plyrPlayer.plyr.embed.getDuration();
-                    if (videoDuration === 0) {
-                        return;
-                    }
                     if (!isNaN(resumeTime) && resumeTime < videoDuration - 3) {
                         plyrPlayer.plyr.embed.seekTo(resumeTime);
                     }
@@ -223,13 +220,9 @@ var ZenPlayer = {
             plyrPlayer.addEventListener("timeupdate", function() {
                 // Store the current time of the video.
                 var resumeTime = 0;
-                if (window.sessionStorage) {
+                var videoDuration = plyrPlayer.plyr.embed.getDuration();
+                if (window.sessionStorage && videoDuration > 0) {
                     var currentTime = plyrPlayer.plyr.embed.getCurrentTime();
-                    var videoDuration = plyrPlayer.plyr.embed.getDuration();
-                    if (videoDuration === 0) {
-                        return;
-                    }
-
                     // Only store the current time if the video isn't done
                     // playing yet. If the video finished already, then it
                     // should start off at the beginning next time.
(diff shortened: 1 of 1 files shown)


js/everything.js
```diff
@@ -612,7 +618,7 @@ $(function() {
     $("#form").submit(function(event) {
         event.preventDefault();
         var formValue = $.trim($("#v").val());
-        var formValueTime = /&t=(\d*)$/g.exec(formValue);
+        var formValueTime = /[?&]t=(\d*)$/g.exec(formValue);
```
shakeelmohamed: 👍

js/everything.js
```diff
@@ -612,7 +618,7 @@ $(function() {
...
         if (formValueTime && formValueTime.length > 1) {
             formValueTime = parseInt(formValueTime[1], 10);
             formValue = formValue.replace(/&t=\d*$/g, "");
```
shakeelmohamed: With the above regex change, I'm concerned that we might introduce a bug here when the querystring looks like `?t=10&v=abcd`

js/everything.js
```diff
@@ -193,6 +193,9 @@ var ZenPlayer = {
                 if (window.sessionStorage && window.sessionStorage.hasOwnProperty(videoID)) {
                     var resumeTime = window.sessionStorage[videoID];
                     var videoDuration = plyrPlayer.plyr.embed.getDuration();
+                    if (videoDuration === 0) {
```
shakeelmohamed: Both of these conditions seem strange, can you provide steps to reproduce the bug?

avalan4e57: Here is some console output whithout fix:
screenshot 2017-07-24 07 48 39
You can see that videoDuration equals 0 when some first function calls. So the closure 
```javascript
if (currentTime < videoDuration - 3) {
  resumeTime = currentTime;
}
```
has no chance to set resumeTime to anything but the default zero.

js/everything.js
avalan4e57: Today youtube reference "Copy video URL at current time" looks like `https://youtu.be/hTWKbfoikeg?t=74`. By the way current version of regexp also can't manage querystrings like `?t=10&v=abcd` or `&t=10&v=abcd`. So, do you want to find some way for manual options setup? If so, what patterns are accepted as valid?

js/everything.js
shakeelmohamed: Thanks for the explanation!

avalan4e57: That looks nicer and everything still works correctly. So why not. My code styling ability is still in progress though

js/everything.js
```diff
@@ -223,6 +226,9 @@ var ZenPlayer = {
                 if (window.sessionStorage) {
                     var currentTime = plyrPlayer.plyr.embed.getCurrentTime();
                     var videoDuration = plyrPlayer.plyr.embed.getDuration();
+                    if (videoDuration === 0) {
```
avalan4e57: Sure thing. It's a good place to learn smth about code styling. I'll add both to my next commit as well as fix regexp to `/[?&]t=(\d*)/g`.

shakeelmohamed:
@avalan4e57 if you've tested this locally, I'm happy to merge it. Thank you so much for your contribution in both finding and fixing a bug!

avalan4e57:
I've tested locally before committing and pushing and once again just now after have read the message. Everything works ok. It was my pleasure to work with you on ZAP. In fact that's gonna be my first commit to an open source project. I like the app and would like to continue with contributing.

shakeelmohamed:
@avalan4e57 excellent! Welcome to the org, feel free to poke around at our long backlog of tasks 😄  and I'm always happy to help if you have any questions

PR #256, opened 2017-07-26
Fix video resume

avalan4e57:
### Motivation and Context
The problem was that video always started from the time argument in URL.
Now video starts from where we left off. But updates the time position by a given argument in URL if loaded for the first time during the session or if the form was submitted again.

### Types of changes

### Description
I've read the code and found out that my change is going to help it work as it was meant to according to it's logic. According to the code we want to resume video from where we left off during the session. Previously video always resumed from time argument in URL. I've fixed that. But then found out that if I want to submit form with the same video but some other time argument the app wouldn't do it and simply resume video from where left off. So I've added a line of code to 'form submission' part to make it update session storage value with the time argument from passed URL. So, that's it. I hope now it works correctly together with some accepted previous commits.

### Final checklist:
- [x] All tests passed.

### DIFF
--- a/js/everything.js
+++ b/js/everything.js
@@ -173,7 +173,8 @@ var ZenPlayer = {
                 // Updates the time position by a given argument in URL
                 // IE https://zenplayer.audio/?v=koJv-j1usoI&t=30 starts at 0:30
                 var t = getCurrentTimePosition();
-                if (t) {
+                // If played for the first time during session and there's 't' option in query
+                if (t && window.sessionStorage[videoID] === undefined) {
                     that.videoPosition = t;
                     window.sessionStorage[videoID] = t;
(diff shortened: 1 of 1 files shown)


shakeelmohamed:
@avalan4e57 hey, sorry I haven't taken a look until now. Not sure why the tests are failing on CI, are they passing for you locally?

avalan4e57:
@shakeelmohamed oh, it's good you've found some time for this. I haven't worked on the project because I thought it's not quite dynamic anymore. And now I can't really remember what could go wrong. But I'll check it out as soon as possible. Glad that project is still alive because it's quite awesome.

shakeelmohamed:
Closing due to inactivity

Issue #222, opened 2016-07-04
Autoplay Feature from Youtube


yoshrubin:
I'm looking to add the "Autoplay" feature from Youtube (can have a on/off toggle) I use it all the time when streaming music on Youtube and think it would be a great feature added here. 
A starting point might be [here](https://developers.google.com/youtube/) to check the API and get some sample code.

cipher1729:
The 'Autoplay' parameter in the youtube player API controls whether the video automatically starts playing when it's loaded. Since that's not applicable here,  did you mean the "automatic playing of next video" feature ?  Would anything be available in the API for that  ?

yoshrubin:
Yea, the automatic playing of next is what I want to implement, not sure if there is a API for that havent had a chance to look.

cipher1729:
Ok, interesting feature if it could be integrated. I'll also look into it

lazy-sunshine:
is this issue still open for working?

shakeelmohamed:
@lazy-sunshine sure, see the work that was started in PR #258 by @avalan4e57. It hasn't been touched in months so you can finish it up based on the feedback I left there. Feel free to ping me with any questions

aarondchu:
is this feature still up for grabs? how is the progress going?

shakeelmohamed:
@aarondchu thanks for your interest, this is still up for grabs. Reading #231 and #258 would be a good starting point. Ping me if you have any questions

k----n:
Any updates? Is this up for grabs again?

shakeelmohamed:
@k----n if the issue is open, it's up for grabs

pgoodbread:
I would like to take this, if possible.

lcpunch:
Is this issue still open? Where are you guys at #329 ?

shakeelmohamed:
@lcpunch go for it, there's been no activity on this project lately 👍

k----n:
@lcpunch @shakeelmohamed 

I implemented autoplay the next related video in this PR: https://github.com/zen-audio-player/zen-audio-player.github.io/pull/329

Not sure if it's mergeable?

lcpunch:
@k----n @shakeelmohamed , I just tested it and it works. The only problem that I see  is when the autoplay is activated it "breaks" the line.

![image](https://user-images.githubusercontent.com/6640213/79229704-7a629600-7e31-11ea-9bab-594babac1ff4.png)

[...]

PR #258, opened 2017-07-28
Closes Issue #222: Autoplay Feature from Youtube

avalan4e57:
### Motivation and Context
There's a pull request (#231) but it isn't active for a long time. So I've used that pull request and made some work with it to close the issue #222.
### Types of changes

Cleaned up the code from pull request #231.
    Then fixed some bugs:
    (1) Fixed getNewVideoID function. Previously it just popped up all
    the playList and left only the first element. And if this element
    would be current video id then nextId will be set to null. Now I've
    made a reverse for playlist array to get just the same next video from
    YouTube as Youtube suggests itself and that is done with no further
    changes in the code. I've also removed the while loop from
    getNewVideoID function and that gives the code an opportunity
    to check all the playlist for a next track during the session.
    (2) Fixed the problem when tracks were added to videoMetadata["items"]
    when they were already there. Now there is a clause that checks if
    current track has already been played and if so just doesn't add
    a duplicate.

### Final checklist:
- [x] All tests passed.

### DIFF
--- a/index.html
+++ b/index.html
@@ -74,6 +74,9 @@
                 <button class="btn" id="toggleDescription">
                     Show Description
                 </button>
+                <button class="btn" id="toggleAutoplay">
+                    Autoplay
+                </button>    
                 <input id="toggleRepeat" name="toggleRepeat" type="checkbox" />
                 <label id="toggleRepeatLabel" for="toggleRepeat">Repeat track</label>
             </div>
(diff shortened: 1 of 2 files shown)


avalan4e57:
Hello! I need to know which event is connected to query search. If I put some video id in search field then I can handle it in form submit event. But if it's a query search like when you just type something like "smells like teen spirit" then code from submit form event listener doesn't work. Can you please help find out what is the name of that exact event?

shakeelmohamed:
@avalan4e57 I think you want `getSearchResults()`

shakeelmohamed:
Closing due to inactivity

Issue #227, opened 2016-07-27
Add Ability to Select Playback Quality


artforlife:
This should mimic the Youtube's range of 480 - 1080P of stream quality.

It would also be nice to implement an algorithm that analyzes the actual sound quality. That is, often people upload a CD quality which is 44kHz but the video will be HD, so Youtube will state 1080P. Perhaps we can output bitrate or something along those lines.

shakeelmohamed:
> This should mimic the Youtube's range of 480 - 1080P of stream quality.

Don't forget 144p & 240p 😆 
We'll probably have to implement this in Plyr, luckily there's already an issue for on their repo (https://github.com/Selz/plyr/issues/218). The challenge is, we can't upgrade until https://github.com/Selz/plyr/issues/199 is done because the latest version of Plyr will break our UI. There's some discussion about this in https://github.com/Selz/plyr/issues/189

cc @SamPotts, no rush just FYI for you

> It would also be nice to implement an algorithm that analyzes the actual sound quality. 

I'd be impressed if we can pull this off in pure JS.

artforlife:
Is there a way to get this info using Youtube API? After all, we probably want a switch for the user to select the quality he wants.

shakeelmohamed:
I think so, you'd have to look into the iframe API docs, should be able to find that page with a quick search

vlastapolach:
Hi, I found in 3 sources that since 2014 youtube streams separate audio - for 144p and for 1080p the same.. 

“YouTube streams video and audio separately and the web player/app combines them on the fly. Due to this, the audio bitrate is not directly affected by video quality like in the past.” https://www.h3xed.com/web-and-internet/youtube-audio-quality-bitrate-240p-360p-480p-720p-1080p

“Starting in 2014, YouTube is streaming a 192kbps AAC audio stream regardless of video resolution (assuming the original upload was of this quality or higher).  Prior to 2014 the audio differed by video resolution.” https://www.quora.com/Does-changing-the-video-quality-on-YouTube-also-affect-the-sound-quality-in-any-way?share=1

“To save you some bandwidth, we select the lowest possible video quality, Don't worry, this doesn't affect the audio in any way. Enjoy!” http://headsetapp.co/

shakeelmohamed:
@vlastapolach great find! It would be nice to still use 480p when available to reduce bandwidth consumption

shakeelmohamed:
Closing as this doesn't seem like something worth doing.

PR #272, opened 2017-10-28
Video Quality Selector button

sandeeproy99:
As asked, proposing a video quality selection button, which is a button click event generating a  selection list.


shakeelmohamed:
Thanks @sandeeproy99! I don't know if we're actually able to implement quality selection, see the discussion in #227

sandeeproy99:
Ok sir. I will try to solve another issue.

PR #274, opened 2017-10-31
closes #222 autoplay-feature-from-youtube

### LINKED ISSUE #222: Autoplay Feature from Youtube

lazy-sunshine:

### DIFF
--- a/js/everything.js
+++ b/js/everything.js
@@ -234,7 +234,7 @@ var ZenPlayer = {
                 videoMetadata = getParsedvideoMetadata();
                 var index = -1;
                 if (videoMetadata) {
-                    var has_duplicates = function (arr) {
+                    var hasDuplicates=function (arr) {
 
                         var len = arr.length;
                         for (var i = 0; i < len; i++) {
@@ -245,10 +245,12 @@ var ZenPlayer = {
                         }
                         return -1;
 
-                    }
-                    index = has_duplicates(videoMetadata.items);
+                    };
+                    index = hasDuplicates(videoMetadata.items);
 
-                    if (index != -1) videoMetadata["items"].splice(index, 1);
+                    if (index !== -1) {
+                        videoMetadata["items"].splice(index, 1);
+                    }
                     videoMetadata["items"].push({ "type": "youtube", "id": currentVideoID, "played": true });
                     videoMetadata["autoplayState"] = autoplayState;
                 }
@@ -754,14 +756,14 @@ $(function () {
         highlight: true,
         minLength: 1
     }, {
-            source: function (query, processSync, processAsync) {
+        source: function (query, processSync, processAsync) {
                 getAutocompleteSuggestions(query, function (data) {
                     return processAsync($.map(data[1], function (item) {
                         return item[0];
                     }));
                 });
             }
-        }).bind("typeahead:selected", function (obj, datum) {
+    }).bind("typeahead:selected", function (obj, datum) {
             window.location.href = makeSearchURL(datum);
         });
 
--- a/js/everything.js
+++ b/js/everything.js
@@ -8,6 +8,10 @@ var timeIntervals = {
     SECONDS: 60
 };
 
+// global playlist, this is populated with an ajax call
+var playList = [];
+var autoplayState = false;
+var videoMetadata = {};
 // Pointer to Keen client
 var client;
 /**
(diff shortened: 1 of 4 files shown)


shakeelmohamed:
Hi @lazy-sunshine! There's some code style issues to fix, click the details link next to the Travis CI check above to see the details. You can run `npm install` then `npm test` locally to verify your changes.

lazy-sunshine:
The PR has been generated.Please review

js/everything.js
```diff
@@ -293,23 +339,43 @@ var ZenPlayer = {
...
+        $("#togglePlayer").click(function (event) {
             toggleElement(event, ".plyr__video-wrapper", "Player");
         });
     },
-    getVideoDescription: function(videoID) {
+    setupAutoplayToggle: function () {
+        // toggle auto next song playing
+        $("#toggleAutoplay").click(function (event) {
+            var toggleTextElement = $("#" + event.currentTarget.id);
+            if (!autoplayState) {
```
shakeelmohamed: the if/else logic here can be simplified

js/everything.js
```diff
@@ -8,6 +8,10 @@ var timeIntervals = {
     SECONDS: 60
 };
 
+// global playlist, this is populated with an ajax call
+var playList = [];
```
shakeelmohamed: I don't know if playlist is the right name since this is really a list of upcoming "suggested" videos

lazy-sunshine:
I have refactored the code according to the suggestion.Please review

js/everything.js
```diff
@@ -225,7 +230,44 @@ var ZenPlayer = {
...
+                    var hasDuplicates = function (arr) {
+                        var len = arr.length;
+                        for (var i = 0; i < len; i++) {
+                            if (arr[i].id === currentVideoID) {
+                                return i;
+                            }
+                        }
+                        return -1;
+                    };
+                    index = hasDuplicates(videoMetadata.items);
```
shakeelmohamed: Looks like this function is actually doing 2 things, and should probably be renamed:

1. Return -1 if all items are unique
2. Return the index of the duplicate

js/everything.js
```diff
@@ -225,7 +230,44 @@ var ZenPlayer = {
...
+
+                    if (index !== -1) {
+                        videoMetadata["items"].splice(index, 1);
```
shakeelmohamed: If we splice, we're actually losing track of videos we've already played.
What if we instead store 2 arrays: played videos, and suggested videos.
When we pick the next suggested video, ensure it's not in the played videos list - if it is remove it.
When we query the API for suggested videos, only add the videos if they're not already in either list.

lazy-sunshine: here I am just changing the order of playlist since if duplicate exist then I have to change the order since this playlist was requested again so I put it on the top, so there is no need of one new playlist

js/everything.js
```diff
@@ -562,7 +622,49 @@ function pickDemo() {
...
+    // if playList songs left
+    var nextID = null;
+    while (suggestedPlayList.length > 0) {
+        nextID = suggestedPlayList.pop();
+        for (var i = 0; i < videoMetadata["items"].length && nextID; i++) {
+            var curVideo = videoMetadata["items"][i];
+            if (curVideo["id"] === nextID) {
+                // restart the for loop
+                nextID = suggestedPlayList.pop();
+                i = -1;
```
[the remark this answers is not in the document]
lazy-sunshine: by putting i =-1 I am rechecking each value for the new nextId that I have found

js/zap-common.js
```diff
@@ -81,4 +81,19 @@ function parseYoutubeVideoID(uri) {
 
         return videoInfo;
     }
-}/* eslint-enable */
\ No newline at end of file
+}/* eslint-enable */
+
+function getAutoplayList(currentVideoID, youTubeDataApiKey, suggestedPlayList, successCallback, onFail) {
+    $.ajax({
```
lazy-sunshine: On keeping the function in  zap-common its giving some linting issue

shakeelmohamed:
@divayprakash it's mostly good, just need to cleanup the format of session variables to use multiple variables instead of storing a JSON blob

shakeelmohamed:
Closing due to inactivity

PR #284, opened 2018-02-06
Update Player.js because it was showing small duration on firefox. I …

chaymafourati:
27721241_10213601051414799_1793452721_n
…think that this will fix the bug

I removed from the span class the --duration

### DIFF
--- a/bower_components/plyr/src/js/plyr.js
+++ b/bower_components/plyr/src/js/plyr.js
@@ -767,7 +767,7 @@
                 html.push(
                     '<span class="plyr__time">',
                         '<span class="plyr__sr-only">' + config.i18n.duration + '</span>',
-                        '<span class="plyr__time--duration">00:00</span>',
+                        '<span class="plyr__time">00:00</span>',
                     '</span>'
                 );
             }


shakeelmohamed:
@chaymafourati this is an interesting bug. Can we fix it another way without modifying plyr.js? This ia 3rd party component and I'd prefer to leave that code intact since we may upgrade to a newer version at some point

chaymafourati:
Great! I will try to work on it another way

shakeelmohamed:
Hi @chaymafourati, the file you've edited is still part of the plyr library. Please make any CSS changes in the `css` directory and revert any changes in the `dist` directory.

bower.json
```diff
@@ -20,7 +20,7 @@
     "keen-js": "~3.4.0",
     "trackjs": "~2.1.16",
     "typeahead.js": "~0.11.1",
-    "plyr": "1.6.x",
+    "plyr": "3.2.0",
```
shakeelmohamed: Just updating this file doesn't update the version number, you'd also need to run `bower install`.
BUT, we don't want to update the version of `plyr` because the theme in newer will break our layout since the controls show the background of the video behind them.

bower_components/plyr/dist/plyr.css
```diff
@@ -1 +1,599 @@
...
\ No newline at end of file
+.plyr .plyr__video-embed iframe,.plyr__tooltip {
```
shakeelmohamed: changes to this file need to be reverted

bower.json
chaymafourati: Okay, i will try to find another work around or if you think this issue cannot be resolved, should it be closed ?

bower_components/plyr/src/js/plyr.js
```diff
@@ -767,7 +767,7 @@
                 html.push(
                     '<span class="plyr__time">',
                         '<span class="plyr__sr-only">' + config.i18n.duration + '</span>',
-                        '<span class="plyr__time--duration">00:00</span>',
+                        '<span class="plyr__time">00:00</span>',
```
shakeelmohamed: this file still needs to be reverted

bower_components/plyr/dist/plyr.css
```diff
@@ -1 +1 @@
...
...vertical-align:middle;font-size:14px;line-height:.95}.plyr__time+.plyr__time{display:none}@media (min-width:768px){.plyr__time+.plyr__time{display:inline-block}}.plyr__time+.plyr__time::before{content:'\2044';margin-right:10px}.plyr__volume{display:none}.plyr .plyr__volume{-ms-flex:1;flex:1;position:relative}.plyr .plyr__volume input[type=range]{position:relative;z-index:2}@media (min-width:480px){.plyr .plyr__volume{display:block;max-width:60px}}@media (min-width:768px){.plyr .plyr__volume{max-width:100px}}.plyr--is-ios .plyr__volume,.plyr--is-ios [data-plyr=mute]{display:none!important}.plyr--fullscreen,.plyr--fullscreen-active{position:fixed;top:0;left:0;right:0;bottom:0;height:100%;width:100%;z-index:10000000;background:#000;border-radius:0}.plyr--fullscreen video,.plyr--fullscreen-active video{height:100%}.plyr--fullscreen .plyr__video-wrapper,.plyr--fullscreen-active .plyr__video-wrapper{height:100%;width:100%}.plyr--fullscreen .plyr__controls,.plyr--fullscreen-active .plyr__controls{position:absolute;bottom:0;left:0;right:0}.plyr--fullscreen-active.plyr--vimeo .plyr__video-wrapper,.plyr--fullscreen.plyr--vimeo .plyr__video-wrapper{height:0;top:50%;transform:translateY(-50%)}
```
shakeelmohamed: this file needs to be reverted

bower.json
shakeelmohamed: @chaymafourati the fix should be a pure CSS change.
I would update `css/styles.cs` with a new rule, something like

```
.plyr__time {
  /** Your style rules here **/
}
```

If it doesn't take effect, add `!important` for now then we can work on getting the fix in without using it.

PR #287, opened 2018-03-04
Removes unnecessary CSSLint warnings from CodeClimate

divayprakash:
### Motivation and Context
	  Removes unnecessary CSSLint warnings from CodeClimate

### Types of changes

### Final checklist:
- [ ] My code follows the code style of this project.
- [ ] All tests passed.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -2,6 +2,9 @@
 engines:
   csslint:
     enabled: true
+    checks:
+      ids:
+        enabled: false
   duplication:
     enabled: true
     config:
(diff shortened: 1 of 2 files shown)


divayprakash:
Fixes 35 CSSLint issues under the two heads -
* Rule doesn't have all its properties in alphabetical order
	* This is actually fixed by re-ordering the CSS properties
* Don't use IDs in selectors
	* This rule has been widely regarded as being immaterial in the case of small web-pages, so I fixed the same by simply removing the check in CSSLint engine configuration in `.codeclimate.yml`

shakeelmohamed:
Thanks @divayprakash

PR #304, opened 2018-10-29
commented h2 element

kapeyka:
h2 element on line 44 doesn't have a header, so it was commented to not execute.


shakeelmohamed:
Thanks for your PR @kapeyka, this `h2` is filled in dynamically by JavaScript if the user is running a mobile device. See https://github.com/zen-audio-player/zen-audio-player.github.io/blob/396b4f613c5dff4f98050419118bab13c7c3c71d/js/everything.js

PR #322, opened 2019-06-05
Add support for 'time_continue' query string parameter

### LINKED ISSUE #301: Add support for time_continue querystring value
`time_continue` should be treated the same as the `t` querystring value.

Example URL format:

`https://www.youtube.com/watch?time_continue=137&v=XYZ`

This format seems to be used with embedded videos.

hchhabra:
### Motivation and Context
- [x] Why is this change required? What problem does it solve?
It adds support of 'time_continue' query string parameter. Fixes #301 .
- [x] If it fixes an open issue, include the text `Closes #1` (where 1 would be the issue number) to your commit message.
Done.

### Types of changes

### Description
Add support for 'time_continue' query string parameter. 'time_continue' parameter behaves same as already supported 't' parameter.

### Final checklist:
- [x] All tests passed.

### DIFF
--- a/js/everything.js
+++ b/js/everything.js
@@ -586,9 +586,8 @@ $(function() {
         var formValue = $.trim($("#v").val());
         var formValueTime = /[?&](t|time_continue)=(\d+)/g.exec(formValue);
         if (formValueTime && formValueTime.length > 2) {
+            formValue = formValue.replace(formValueTime[0], "");
             formValueTime = parseInt(formValueTime[2], 10);
-            formValue = formValue.replace(/[?&]t=\d*/g, "");
-            formValue = formValue.replace(/[?&]time_continue=\d*/g, "");
         }
         if (formValue) {
             var videoID = wrapParseYouTubeVideoID(formValue, true);
--- a/js/everything.js
+++ b/js/everything.js
@@ -584,16 +584,11 @@ $(function() {
     $("#form").submit(function(event) {
         event.preventDefault();
         var formValue = $.trim($("#v").val());
-        var formValueTime = /[?&]t=(\d+)|[?&]time_continue=(\d+)/g.exec(formValue);
+        var formValueTime = /[?&](t|time_continue)=(\d+)/g.exec(formValue);
         if (formValueTime && formValueTime.length > 2) {
-            if (typeof formValueTime[1] !== "undefined") {
-                formValueTime = parseInt(formValueTime[1], 10);
-            }
-            else {
-                formValueTime = parseInt(formValueTime[2], 10);
-            }
-            formValue = formValue.replace(/[?&]t=\d+/g, "");
-            formValue = formValue.replace(/[?&]time_continue=\d+/g, "");
+            formValueTime = parseInt(formValueTime[2], 10);
+            formValue = formValue.replace(/[?&]t=\d*/g, "");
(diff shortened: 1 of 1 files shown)


js/everything.js
```diff
@@ -372,9 +372,13 @@ function getCurrentVideoID() {
  */
 function getCurrentTimePosition() {
     var t = parseInt(URI(window.location).search(true).t, 10);
+    var timeContinue = parseInt(URI(window.location).search(true).time_continue, 10);
     if (t > 0 && t < Number.MAX_VALUE) {
         return t;
     }
+    else if (timeContinue > 0 && timeContinue < Number.MAX_VALUE) {
```
shakeelmohamed: this change is fine

js/everything.js
```diff
@@ -580,10 +584,16 @@ $(function() {
     $("#form").submit(function(event) {
         event.preventDefault();
         var formValue = $.trim($("#v").val());
-        var formValueTime = /[?&]t=(\d*)/g.exec(formValue);
-        if (formValueTime && formValueTime.length > 1) {
-            formValueTime = parseInt(formValueTime[1], 10);
-            formValue = formValue.replace(/&t=\d*$/g, "");
+        var formValueTime = /[?&]t=(\d+)|[?&]time_continue=(\d+)/g.exec(formValue);
+        if (formValueTime && formValueTime.length > 2) {
```
shakeelmohamed: I think we can simplify this regex to`/[?&](t|time_continue)=(\d*)/g`. The difference with mine is we say "we'll take either `t` or `time_continue` and use the first one we find". Doing this should allow us to simplify the logic below

Play with it: https://regex101.com/r/VaRFn0/3

hchhabra: Sure, looks good, @shakeelmohamed . Will use that one. One thing is that I think it should be `/[?&](t|time_continue)=(\d+)/g`, i.e. `\d+` instead of `\d*` to match atleast one digit following the `=` sign, so that it works for cases like `?time_continue=&t=2` properly. Of course those are invalid cases, but we may want to handle them correctly?

shakeelmohamed: @hchhabra oh yes you're right about `\d+`, I was playing with that regex quickly. Go for it!

hchhabra: Made the change, @shakeelmohamed .

js/everything.js
```diff
@@ -580,10 +584,11 @@ $(function() {
...
         event.preventDefault();
         var formValue = $.trim($("#v").val());
-        var formValueTime = /[?&]t=(\d*)/g.exec(formValue);
-        if (formValueTime && formValueTime.length > 1) {
-            formValueTime = parseInt(formValueTime[1], 10);
-            formValue = formValue.replace(/&t=\d*$/g, "");
+        var formValueTime = /[?&](t|time_continue)=(\d+)/g.exec(formValue);
+        if (formValueTime && formValueTime.length > 2) {
+            formValueTime = parseInt(formValueTime[2], 10);
+            formValue = formValue.replace(/[?&]t=\d*/g, "");
```
shakeelmohamed: We can be a bit more selective here, and use `formValueTime[1]` (which will be either `t` or `time_continue`) so we don't have to do the replacement twice

hchhabra: OK, I thought the input could possibly contain both. If that's not the case, I'll change it.

PR #323, opened 2019-07-18
Closes #321 disable autoplay and execute show() on ready flag

### LINKED ISSUE #321: Functionality is totally broken
I'm not sure exactly what's happened, but I think YouTube has made some API changes around CORS with the player widget.

TODO:
- [ ] Investigate & repurpose this issue

andy-shi88:
Disable autoplay due to the new autoplay policy on some browser.

### Important

### Motivation and Context
- [x] Since browser now don't support autoplay by default (it require user to has some action taken first), and we only execute the `show()` function on `playing` flag which is triggered by the autoplay, the player will never show.
- [x] Closes #321 
- [x] autoplay is disabled now.

### Types of changes
- [x] Breaking change (fix or feature that would cause existing functionality to change), not sure if it's count as breaking or not, but have to disable autoplay.

### Description
- [ ] disable autoplay to follow the new autoplay policy
- [ ] execute show() in `ready`

### Final checklist:
- [x] All tests passed.


js/everything.js
```diff
@@ -137,6 +136,7 @@ var ZenPlayer = {
                 that.setupTitle();
                 that.setupVideoDescription(videoID);
                 that.setupPlyrToggle();
+                that.show();
```
shakeelmohamed: Showing the player defeats the purpose of this project, can we instead start playing here? I had that working once upon a time, then reverted it once we set `autoplay=true`

shakeelmohamed:
Thanks for getting this started @andy-shi88! 🎉

js/everything.js
andy-shi88: we can start playing here by calling `play`, but still we'll need the user to enable the `autoplay` on the site already to do that, we'll get `Autoplay is only allowed when approved by the user, the site is activated by the user, or media is muted.`  

one way we can do it is, we can `play` here, but in the `play` function we'll need to check if autoplay is not allowed by user, then we show the player to let user click the play button.

andy-shi88:
@shakeelmohamed I pushed new changes that play on ready, but check in 2 second if the status is still not playing then consider the autoplay is being disabled, then show the player.

Is there any other way to detect the embedded player fail to play? in `audio` or `video` it's possible to check if `play()` function is being hold by the autoplay flag [here](https://stackoverflow.com/a/49939655/7616247) but for embedded player I can't find any, that's why I go with 2 second check instead.

andy-shi88:
@shakeelmohamed can you check?

shakeelmohamed:
@andy-shi88 Can we do something like show the player, play the video, then hide the player again?

shakeelmohamed:
Closing due to inactivity.

PR #332, opened 2019-10-23
Closes #331 Fixed focus button click event

### LINKED ISSUE #331: Clicking near focus button redirects to another url
Trying the website https://zenplayer.audio/ I've found that there is an event that triggers when the user clicks on the div element instead of doing that only on the focus button. To test it you just have to click near the focus button to see how the url is redirected to what is supposed to be the url in the focus button.

I've tried it with the latest version of both Firefox and Chrome and the result is the same.

abnazhor:
Fixed wrong positioning of id on focus button.

### DIFF
--- a/index.html
+++ b/index.html
@@ -93,9 +93,9 @@
             Click for Demo
         </button>
 
-        <div id="focus-btn">
+        <div>
           <br>
-          <button class="btn">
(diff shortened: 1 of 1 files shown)


shakeelmohamed:
Thanks @abnazhor!

Issue #335, opened 2019-10-24
Sponsors isn't adjusted at 760x800 resolutions.


abnazhor:
Currently sponsors are correctly aligned in all cases except when the resolution is around 760px. It looks something like this:
![imagen](https://user-images.githubusercontent.com/11428398/67528284-611a3180-f6b9-11e9-8723-4d506694157b.png)

abnazhor:
Also, I would love to work on this issue. Can I?

shakeelmohamed:
@abnazhor please go ahead

abnazhor:
@shakeelmohamed Done!

shakeelmohamed:
Closed with #338

PR #338, opened 2019-10-28
#335 Sponsors' position fixed

abnazhor:
Fixes error with positioning in issue #335 
- Removed media query to replace it with an alternative that works on any screen resolution.
- Fixed positioning using flexbox


shakeelmohamed (approved):
Oops, totally forgot about this. Thanks @abnazhor!

PR #371, opened 2021-07-21
closes #368 amending changes requested

ssharvin:
-Done the following changes.
Omitting the Facebook and Twitter social media icons.
Properties at CSS for the social media icons are alphabetized.
Issues on line breaking for the coding style is amended.
Duplication of GitHub account source link as it appears in both icon and text message is solved.
#368 
image

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -29,6 +29,36 @@ footer {
     padding: 50px 0;
 }
 
+/* social media icons */
+
+footer .fa {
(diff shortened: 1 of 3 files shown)


shakeelmohamed:
Hi @ssharvin, I'm not sure why there's been multiple PRs for these changes.
This PR looks correct, but I'd like to see these changes in #368 so the feedback and changes are  all in the same PR.

PR #377, opened 2021-10-15
Add rel "preconnect" to font-awesome CDN

Dragonashes:
### Motivation and Context
This change reduces loadtime by sending a earlier request to CDN resource.
### Types of changes
This should be non-breaking.
### Description
This change adds preconnect to the link stylesheet to speed up font downloading, the font is downloaded externally from a ContentDeliveryNetwork. the rel="" attribute accepts a space seperated list for this.
### Final checklist:
Codestyle is followed
I read the contributing.md file
The tests failed for some reason, not caused by me.

PR #378, opened 2021-10-16
rm font-awesome CDN

Dragonashes:
### Motivation and Context
This is an alternative to #377

### Description
This change removes the font-awesome CDN, the icons still show when running the page, this is a small change so either solution would do, it is easy to reintroduce the external source at a later time as it is one line of code.


shakeelmohamed:
Awesome, thanks @Dragonashes!

PR #386, opened 2021-12-30
fix(overflow): fix overflow with TrackJS logo

Omriruvio:
image size was overflowing on smaller resolutions (under 500px)
change image size to maximum 100% (maximum screen width)

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -281,6 +281,10 @@ a {
     overflow: hidden;
 }
 
+.img-40 {
+    width: 100%;
+}
+
 .img-100 {
     display: block;
     margin-left: auto;


css/styles.css
```diff
@@ -281,6 +281,10 @@ a {
     overflow: hidden;
 }
 
+.img-40 {
```
shakeelmohamed, line 284: Thanks for the PR!

This class name is not descriptive of its behavior, though I think we have some other technical debt causing this issue.

Omriruvio: I didn't want to interfere  with other possible JS or other modules relying on this selector.
My suggested change is a simple adjustment to the image size to make sure it doesn't overflow on small screen resolutions.

PR #414, opened 2023-05-01
muzyk

hanonine:


shakeelmohamed:
Closing due to a) no context, b) arbitrary ZIP file committed to the repo which I assume to be malicious, c) a pattern of this user following the same behavior on multiple repos.

PR #416, opened 2023-07-30
Migrate all tests to puppeteer, remove zombie test dependency

shakeelmohamed:
There is likely a breaking change in mocha since v7 which is causing the tests to hang.
One of those changes is the deprecation of `mocha.opts`, though there may be other issues to address.

CI will be good to go once the `workflow` branch is merged in, but CI itself will hang until this issue if fixed.

A possible solution is replacing zombie with [Puppeteer](https://pptr.dev/) for headless testing


OpherM:
Is there any issue to be resolved?

shakeelmohamed:
@OpherM Hello, yes. I just reproduced the hang by running `npm test` locally on Node.js 16.20.1

Issue #433, opened 2024-05-28
Search Results showing Max of 5 Results via Search


maiteemate:
When accessing this URL:  'https://zen-audio-player.github.io/'
the web app is limited to show the Max Results of '5' results when searching for any song or video.

Is there a way to increase this via a setting or configuration file?  This hugely limits what can be played with the web interface on the URL above.

Please let me know if I'm missing anything.
In any event, thank you for a very simple, and useful YT web app.

shakeelmohamed:
Hi @maiteemate!

The [YouTube API](https://developers.google.com/youtube/v3/docs/search/list) defaults to a value of 5 for `maxResults`, so we’re just consuming their default.

It’s not ideal, but as a workaround you can search on youtube.com then copy/paste a link over.

If you’re interested in making a PR for this feature, I’m happy to guide you through it.

RLalor:
A PR to increase the results you mean? Do you want to change it or stick with 5? How would you like it? This seems like an easy edit

shakeelmohamed:
@RLalor I believe youtube.com fetches results in batches of 10, so a quick change is to set `maxResults` to 10.

A more nuanced solution would be loading multiple batches of results.
Here are some possible approaches, ranked most to least preferred (by me):
1. A “load more” button
2. Pagination
2. Infinite scroll like youtube.com

Happy to discuss this further if you’re interested in working on this.

RLalor:
Thanks for the reply. I cloned it and if I end up playing with it and completing something I will let you know.

shakeelmohamed:
@maiteemate I’m closing this for now as we can get 10 results per search now. If you want to see a different solution, please reopen and let’s continue the discussion.

Thanks to @RLalor the quick PR #436.

PR #436, opened 2024-07-10
#433 improved fetch results to 10

RLalor:
## Description
added 1 line of code to increase results to 10 as requested in #433 until one of the other options is implemented.

## Checklist
- [] I have read and followed the project's contributing guidelines.
- [] I have made necessary updates to the documentation.
image

## Summary by Sourcery

Increased the maximum number of fetch results to 10 in the YouTube search query.

### DIFF
--- a/js/zap-common.js
+++ b/js/zap-common.js
@@ -6,6 +6,7 @@ function getSearchResults(query, youTubeDataApiKey, onData, onFail) {
         key: youTubeDataApiKey,
         part: "snippet",
         q: query,
+        maxResults: 10,
         type: "video"
     }, onData).fail(onFail);
 }


shakeelmohamed (approved):
Looks great, thanks @RLalor!

RLalor:
Thanks. This is a nice project to work on. The other changes you mention look quite easy as well. If I get bored I will play with and test those as well and discuss which option is most suited. 
I tested it and it can fetch 50 results per page. If you try 51 it doesn't produce an error but just ignores it and gets 50. I think you can do 50 per page up to 1000 results / 20 pages. But I also read 500.  
Perhaps adding 2 more columns to prevent the page becoming so long is worth considering as well.

